PR #1, opened 2016-01-29
Pretty print JSON on development environments

coreh:
This patch makes it so that the JSON output returned by the server gets pretty-printed whenever `NODE_ENV=development`.


rauchg:
😍

PR #3, opened 2016-01-30
Fix execution instructions

dschenkelman:
The current order results in:

```
➜  micro-sample  micro -p 3000 index.js 
module.js:340
    throw err;
    ^

Error: Cannot find module '/{path}/3000'
```


rauchg:
The former should work though :\ Maybe commander.js being misused?
Thanks!

rauchg:
Fixed in `1.0.1`.

PR #7, opened 2016-02-01
Fixing code continuity error.

ccutch:
Parsing error attributes in function declaration to match `run` and `json` functions


rauchg:
Beautiful. Thanks!

PR #14, opened 2016-02-28
Optimize `JSON.stringify` performance in V8.

bnjmnt4n:
Older versions of V8 have a bug [1](https://bugs.chromium.org/p/v8/issues/detail?id=4730) where using empty `replacer` or `space` arguments decreases performance.


rauchg:
:O!

PR #21, opened 2016-04-08
Refer to zeit/test-listen instead of test/_listen

onbjerg:
This might be a better idea since people don't need to do a copy/pasta to get the helper function.

Maybe we should also use `test-listen` in `micro` itself instead?


rauchg:
We should @onbjerg, I wrote it based on this experience. PR welcome. Thank you !

PR #24, opened 2016-04-28
Use separate micro-serve module

leo:
Like discussed internally with @rauchg.


leo:
I have no idea why, but some tests are failing. Can you check that, @rauchg?

PR #29, opened 2016-05-03
Add main library to package.json

### LINKED ISSUE #28: using `micro file.js` on readme example doesn't work
for recollection the readme example is

``` javascript
import { send } from 'micro';
import sleep from 'then-sleep';
export default async function (req, res) {
  await sleep(500);
  send(res, 200, 'Ready!');
}
```

A solution to this would be to either
- replace with `import { send } from 'micro-core'`
- reexport micro-core from micro (which has no `main` field in its package)

julianduque:
Currently you can't require `micro` because it doesn't find the library.

Fixes #28

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
       "build/test/*.js"
     ]
   },
+  "main": "build/lib/index.js",
   "scripts": {
     "gulp": "gulp",
     "test": "gulp test"


rauchg:
:O

rauchg:
Gracias @julianduque

julianduque:
@rauchg <3

PR #34, opened 2016-05-10
Switch to args and clean up

leo:
This PR introduces the args implementation and tweaks a few other spots in the code (especially the tests and the transpiling workflow).


bin/micro
```diff
@@ -12,25 +12,24 @@ import transformAsync from 'babel-plugin-transform-async-to-generator'
 import transformRuntime from 'babel-plugin-transform-runtime'
 import alias from 'babel-plugin-module-alias'
 
-program
-.version(version)
-.option('-p, --port <port>', 'Port to listen on (3000)', parseInt)
-.option('-n, --no-babel', 'Skip Babel transformation')
-.parse(process.argv)
+args
+  .option('port', 'Port to listen on', 3000, parseInt)
```
rauchg: how about the `-p` alias

leo: @rauchg By default, args uses the first letter of the long name to generate the short name. If you don't want that, you can provide an array instead: `['p', 'port']`

rauchg: the array would be more clear.

leo: @rauchg Adjusted in a61fd6bae844a9c69daeff701814be67d7fb4707

test/index.js
```diff
@@ -1,9 +1,21 @@
 import test from 'ava'
 import { json, send } from 'micro-core'
 import request from 'request-promise'
-import listen from './_listen'
+import micro from 'micro-core'
 import sleep from 'then-sleep'
 
+const listen = async (fn, opts) => {
```
rauchg: `async` not needed since we return a `Promise` and we don't use `await`

leo: Ahh... Removed!

leo:
@rauchg Anything else that needs to be fixed?

rauchg:
This is one of the best PRs ever received. Thanks @leo

leo:
I'm glad it helped :blush:

PR #39, opened 2016-06-03
Added travis ci support for node 5 and 6

amilajack:


rauchg:
Beautiful. Thanks @amilajack

PR #46, opened 2016-06-07
Add buffer and stream support to `send`

kevin-roark:
Once a release of [micro-core](https://github.com/zeit/micro-core) is made, I'll update the dependency here. Tests will fail until then.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "gulp-ext": "1.0.0",
     "gulp-task-listing": "1.0.1",
     "request-promise": "2.0.0",
+    "resumer": "0.0.0",
     "then-sleep": "1.0.1"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "gulp-ext": "1.0.0",
     "gulp-task-listing": "1.0.1",
     "request-promise": "2.0.0",
+    "resumer": "0.0.0",
```
rauchg, line 67: `0.0.0` :O

test/index.js
```diff
@@ -129,6 +174,25 @@ test('throw (500)', async t => {
   }
 })
 
+test('send(200, <Stream>) with error', async t => {
+  const fn = async (req, res) => {
+    const stream = resumer().queue('error-stream')
+    send(res, 200, stream)
+
+    stream.emit('error', new Error('500 from test (expected)'))
```
rauchg, line 182: would be cool to emit it on a different tick just for 💩 and giggles

kevin-roark, line 182: aye-aye :)

Issue #48, opened 2016-06-09
Cannot run micro 4.0.0


julianduque:
```
⬢  microtest  micro -p 7001 one.js
module.js:327
    throw err;
    ^

Error: Cannot find module 'babel-plugin-syntax-async-functions'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/julianduque/.nvm/versions/node/v4.4.5/lib/node_modules/micro/build/bin/micro:20:40)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
```

julianduque:
It seems module published to npm has incorrect build output

```
⬢  micro [master] diff build/bin/micro ~/.nvm/versions/node/v4.4.5/lib/node_modules/micro/build/bin/micro
38c38
< _commander2.default.version(_package.version).option('-p, --port <port>', 'Port to listen on (3000)', parseInt).option('-h, --host <host>', 'Host to listen on (0.0.0.0)').option('-n, --no-babel', 'Skip Babel transformation').parse(process.argv);
---
> _commander2.default.version(_package.version).option('-p, --port <port>', 'Port to listen on (3000)', parseInt).option('-n, --no-babel', 'Skip Babel transformation').parse(process.argv);
75d74
< var host = _commander2.default.host || '0.0.0.0';
77c76
< (0, _microCore2.default)(mod).listen(port, host, function (err) {
---
> (0, _microCore2.default)(mod).listen(port, function (err) {
83c82
<   console.log('> \u001b[96mReady!\u001b[39m Listening on ' + host + ':' + port + '.');
---
>   console.log('> \u001b[96mReady!\u001b[39m Listening on ' + port + '.');
```

rauchg:
@julianduque `4.1.0` should be good!

julianduque:
<3

PR #49, opened 2016-06-09
Add prepublish command to npm scripts

julianduque:
Never forget to build before a publish, addresses #48


rauchg:
This is so useful it almost should be mandatory to have `prepublish`. Thanks @julianduque

PR #52, opened 2016-06-11
Use `main` as entry point for application if file isn't passed as argument

### LINKED ISSUE #47: CLI should take entry point file from package.json#main if not provided as argument
That would make my `npm start` script slightly more micro.

julianduque:
It will look for `main` property in package.json and will try to use it as main file when argument isn't provided.

Fixes #47

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "gulp-ext": "1.0.0",
     "gulp-task-listing": "1.0.1",
     "request-promise": "2.0.0",
-    "resumer": "0.0.0",
     "then-sleep": "1.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,7 +64,6 @@
     "gulp-ext": "1.0.0",
     "gulp-task-listing": "1.0.1",
     "request-promise": "2.0.0",
-    "resumer": "0.0.0",
```
rauchg, line 67: I thought we were using this in tests?

julianduque:
@rauchg you added it here on dependencies: https://github.com/zeit/micro/commit/01171e1291e4bc59c6873652c430e6b4f9e641e1, so it's duplicated, so it should be on deps or devDeps?

rauchg:
`devDeps`, my bad

rauchg:
Awesome contributions @julianduque

Issue #56, opened 2016-06-15
host option conflicts with help flag


millette:
Oups...

```
micro -h localhost index.js # displays help
micro --host localhost index.js # works as expected
```

rauchg:
ugh :\ let's do `-H` ?

millette:
https://www.npmjs.com/package/httpserver uses -H but I've also seen -b (bind). I'll fix it either way, let me know your preference.

julianduque:
`-H` sounds good

PR #57, opened 2016-06-17
Use -H for host flag

millette:
Fix for #56


rauchg:
Thanks @millette for all your amazing contributions!

Issue #58, opened 2016-06-26
file argument is ignored when no-babel option is used


nkzawa:
```
micro -n index.js

> Error! Please supply a file.
```

It seems the file argument is mistakenly interpreted as the value of `no-babel` option.

julianduque:
@nkzawa can you test with current `master` if you are still seeing the same error?

nkzawa:
I can't reproduce with current master.

julianduque:
@nkzawa so it works with current `master`? Lets wait until @rauchg publishes a new version to npm \o/, thanks for your report.

leo:
`v5.0.0` was released 6 days ago. @nkzawa does the bug still occur with that release?

nkzawa:
it has already been fixed

PR #59, opened 2016-06-26
improve error handling on cli

julianduque:
Possibly related to #58 

Currently `micro` isn't handling an error if the loaded files fails for some reason, only error handled is when file isn't found.


rauchg:
Thanks! Good (puts shades on) catch!

PR #71, opened 2016-08-05
Add docs for `createError`

onbjerg:
Per @leo's [request](https://github.com/zeit/micro-core/pull/12#issuecomment-237871452) 🎉


leo:
Thanks a lot! ❤️

PR #78, opened 2016-08-13
Bump async-to-gen to v1.0.3

leebyron:
Significant bug fixes between v1.0.1 and v1.0.3

https://github.com/leebyron/async-to-gen/releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "xo": "^0.16.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


leo:
Thank you for that, but we actually have @greenkeeperio-bot enabled! 😊 So the next time, they'll simply take care of bumping the version.

leebyron:
Slick!

PR #99, opened 2016-10-01
Allows micro cli to use custom onError.

rickharrison:
The micro cli now will read in `exports.onError` and pass it to `serve` if it exists. Here is an example:

``` javascript
module.exports = exports = async function (request, response) {
  throw createError(405, 'Method Not Allowed');
};

exports.onError = function (req, res, { statusCode, message, stack }) {
  send(res, statusCode, {
    statusCode,
    message
  });
}
```

Please let me know if this is something you would be open to merging. If so, I will add tests and documentation. Thanks!

### DIFF
--- a/bin/micro
+++ b/bin/micro
@@ -55,9 +55,11 @@ if ('/' !== file[0]) {
 require('async-to-gen/register')
 
 let mod
+let onError
 
 try {
   mod = require(file);
+  onError = mod.onError
   if (mod && 'object' === typeof mod) {
     mod = mod.default;
   }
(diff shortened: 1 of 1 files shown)


leo:
Thanks!

rickharrison:
Oops, I didn't write the tests yet. I'll write the tests this weekend and submit another PR.

rauchg:
There's a more elegant solution for this: just set up your own try/catch at the top level.

``` js
const withErrors = require('my-app-on-error')
module.exports = withErrors((req, res) => {
  throw new Error('voila!')
})
```

In fact, @nkzawa has brought up before that `onError` is a bit of a stink, and I'm all for deprecating it in the future. Fewer features.

leo:
@rauchg Got it! Let's remove it...

rickharrison:
Cool, that sounds good. I just published `micro-json-error` as a small package for my use case using the method you described: https://github.com/rickharrison/micro-json-error

rauchg:
@rickharrison that's really cool!

Issue #102, opened 2016-10-08
Deprecate `onError`


leo:
https://github.com/zeit/micro/pull/99#issuecomment-252392427

leo:
Fixed in https://github.com/zeit/micro/pull/103#pullrequestreview-3416958

PR #103, opened 2016-10-09
Deprecates `onError` and removes references from README.

rickharrison:
As discussed, in #99 and #102, this adds a console warning message that `onError` is now deprecated.

We could also remove the additions from #99 as no one is probably using them yet? I personally am now using this: https://github.com/rickharrison/micro-json-error

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,13 +74,12 @@ $ npm start
 ### API
 
 #### micro
-**`micro(fn, { onError = null })`**
+**`micro(fn)`**
 
 - This function is exposed as the `default` export.
 - Use `require('micro')`.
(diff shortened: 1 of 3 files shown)


leo:
> We could also remove the additions from #99 as no one is probably using them yet?

Yep, let's do that. 😊

leo (changes requested):
> We could also remove the additions from #99 as no one is probably using them yet?

Yep, let's do that. 😊

rickharrison:
Cool, sounds good. Sorry for causing the churn haha :grin:

leo:
Perfect. Thank you! 🚀 ⭐️

Issue #33, opened 2016-05-09
Visualising requests


leo:
As suggested by @rauchg in #15, we should incorporate support for visualizing the requests in terms of the data flow, with built in JSON highlighting and prettifying, among other things.

Here's an example:

![1b624cf2-e120-11e5-977a-d6508f8a19b0](https://cloud.githubusercontent.com/assets/6170607/15124818/0af4b6f4-162a-11e6-9d99-27c660c19f64.png)

onbjerg:
Hmm. What about sensitive information in requests?

rauchg:
@onbjerg this is primarily intended for development, not production

onbjerg:
@rauchg Oh, of course.

onbjerg:
What about headers and non-JSON requests/responses?

fdidron:
Would be great !
How would one optin to that ? Adding a parameter `--debug` running or simply by having `NODE_ENV !== 'production'` ?

timneutkens:
@fdidron see #104. Adds a `--log` flag for that exact behaviour.

leo:
https://github.com/zeit/micro-dev

onbjerg:
@leo You guys keep linking to that, but the repository is still empty ;)

leo:
@onbjerg @timneutkens is on it. It will go online after the next major release (we'll do that today).

PR #104, opened 2016-10-09
Visualises requests when in development.

rickharrison:
I took a stab at visualising requests from #33. The one thing that is a little messy is parsing and displaying the json. I ended up keeping a reference to it on the request since you can only call `getRawBody` once per request.

Let me know your thoughts!

screen shot 2016-10-08 at 11 51 34 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,16 @@ function handleErrors (fn) {
 }
 ```
 
+### Logging
+
+Micro provides built-in logging to help visualise requests. When `NODE_ENV` is set to `development`, micro will display the request along with any json body like this:
+
+![Development Logging](https://cloud.githubusercontent.com/assets/22048/19218565/2c6081ba-8db2-11e6-93df-9e256a8354fb.png)
+
+This behavior can be overridden by using the cli argument `--log` which can be set to `dev`, `prod`, or `off`. The `prod` version is a condensed log that will not show the json body:
+
+<img width="261" alt="Production Logging" src="https://cloud.githubusercontent.com/assets/22048/21488276/7fba8ea0-cb92-11e6-8b8c-c2f280026630.png">
+
 ### Testing
 
 Micro makes tests compact and a pleasure to read and write.
--- a/bin/micro
+++ b/bin/micro
@@ -29,8 +29,8 @@ const help = () => {
   console.log(`Usage: micro [opts] <file>
   -H, --host  Host to listen on   [0.0.0.0]
   -p, --port  Port to listen on      [3000]
-  -h, --help  Show this help message
-  -L, --log dev|prod|off`);
+  -L, --log   dev|prod|off            [off]
+  -h, --help  Show this help message`);
 }
 
 if (args.h) {
--- a/bin/micro
+++ b/bin/micro
@@ -28,7 +28,8 @@ const help = () => {
   console.log(`Usage: micro [opts] <file>
   -H, --host  Host to listen on   [0.0.0.0]
   -p, --port  Port to listen on      [3000]
-  -h, --help  Show this help message`);
+  -h, --help  Show this help message
+  --log dev|prod|none`);
 }
 
 if (args.h) {
@@ -79,8 +80,8 @@ if ('function' !== typeof mod) {
   process.exit(1)
 }
 
-const { port, host } = args
-serve(mod).listen(port, host, err => {
+const { port, host, log } = args
+serve(mod, {log}).listen(port, host, err => {
   if (err) {
     console.error('micro:', err.stack)
     process.exit(1)
--- a/bin/micro
+++ b/bin/micro
@@ -14,7 +14,8 @@ const args = parse(process.argv, {
   alias: {
     H: 'host',
     h: 'help',
-    p: 'port'
+    p: 'port',
+    L: 'log'
   },
   boolean: ['h'],
   default: {
@@ -29,7 +30,7 @@ const help = () => {
   -H, --host  Host to listen on   [0.0.0.0]
   -p, --port  Port to listen on      [3000]
   -h, --help  Show this help message
-  --log dev|prod|none`);
+  -L, --log dev|prod|off`);
 }
 
 if (args.h) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -63,14 +63,6 @@ async function run(req, res, fn, onError, {log}) {
 
 async function json(req, {limit = '1mb'} = {}) {
   try {
-    if (req.jsonError) {
-      throw req.jsonError
-    }
-
-    if (req.parsedJson) {
-      return req.parsedJson
-    }
-
     const type = req.headers['content-type']
     const length = req.headers['content-length']
     const encoding = typer.parse(type).parameters.charset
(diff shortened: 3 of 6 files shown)


lib/index.js
```diff
@@ -37,6 +55,12 @@ async function run(req, res, fn, onError) {
 
 async function json(req, {limit = '1mb'} = {}) {
   try {
+    if (DEV && req.parsedJson) {
+      return req.parsedJson
+    } else if (DEV && req.jsonError) {
```
timneutkens: Shouldn't we first check if there is an error? 🤔

timneutkens (approved):
Looks fine 👍, maybe add in a way to pass a custom parse function later on, so you can also log out other stuff. urlencoded forms for example.
For now this will do fine 😄

lib/index.js
rickharrison: In theory, there can only be one or the other, but I can switch the order if you like.

rauchg:
This is really beautiful @rickharrison. I think we should have `--log=dev|prod|none` as well that _ignores_ env sniffing.

The prod mode of this would be a _compressed_ version that excludes the body.

rauchg:
(basically something similar to the default express logger)

rickharrison:
Sounds good! I will work on this over my holiday break :)

luisrudge:
This is amazing

rickharrison:
Here is what `--log=prod` looks like:

screen shot 2016-12-26 at 5 40 51 pm

timneutkens (approved):
Looks good :)

timneutkens:
@rauchg final check?

rauchg:
I'll take a look!

bin/micro
```diff
@@ -26,9 +26,10 @@ let [,, file] = args._;
 
 const help = () => {
   console.log(`Usage: micro [opts] <file>
-  -H  Host to listen on   [0.0.0.0]
-  -p  Port to listen on      [3000]
-  -h  Show this help message`);
+  -H    Host to listen on   [0.0.0.0]
+  -p    Port to listen on      [3000]
+  -h    Show this help message
+  --log dev|prod|none`);
```
rauchg: `off` would be better than `none` IMO

rauchg: we also need long forms for all of this, in addition to the short form

```
-H, --host
-p, --port
-h, --help
-L, --log
```

lib/index.js
```diff
@@ -5,8 +5,22 @@ const server = require('http').Server
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
 const isStream = require('isstream')
+const chalk = require('chalk')
+const jsome = require('jsome')
 
 const DEV = 'development' === process.env.NODE_ENV
```
rauchg, line 11: It'd be cool to document in the README that we do the nicer logging upon _explicit_ `NODE_ENV = development` setting

lib/index.js
```diff
@@ -41,6 +63,14 @@ async function run(req, res, fn, onError) {
 
 async function json(req, {limit = '1mb'} = {}) {
   try {
+    if (req.jsonError) {
```
rauchg: I'd like to avoid polluting `req` if at all possible

bin/micro
timneutkens: I'll add your second comment to my readme PR @rauchg

timneutkens: @rauchg PR'ed #122

timneutkens: https://github.com/rickharrison/micro/pull/1

README.md
```diff
@@ -250,6 +250,14 @@ function handleErrors (fn) {
 }
 ```
 
+### Logging
+
+Micro provides some built-in logging to help visualise requests. When `NODE_ENV` is set to `development`, micro will display the request along with any json body like this:
```
timneutkens: Remove `some`

bin/micro
```diff
@@ -28,7 +29,8 @@ const help = () => {
   console.log(`Usage: micro [opts] <file>
   -H, --host  Host to listen on   [0.0.0.0]
   -p, --port  Port to listen on      [3000]
-  -h, --help  Show this help message`);
+  -h, --help  Show this help message
+  -L, --log dev|prod|off`);
```
timneutkens: Turned this around in my merge commit. First the logging then help, also added `[off]` as default indicator

timneutkens:
@rauchg :shipit: 🚀

rauchg:
@rickharrison @timneutkens I think a few additions would make sense:
- Response time (+ delta). EG: `10:30:00 PM (+30ms)`. Kinda like how `debug` gives you deltas
- Also visualize response objects?
- Not 100% sure, but I believe we can anchor the time to the right hand side with an ANSI escape code? (that'd be neat if possible!)

rickharrison:
Sounds good! I'll take a look at this later tonight.

rickharrison:
@rauchg So I have it logging response objects as well now, but I did it by adding a property to `res` 🙁 

The problem is I'm not sure how to get the response body in the `res.on('finish')` event. Is it possible to see the data chunks from a `http.ServerResponse`? An alternative would be to use the callback parameter in the `res.end()` call. However, I do not know what the current log level is in that method. Do you think it would be better to store the log level in a global variable like `DEV` is from the initial `serve` call? Then instead of attaching to `res`, we could output the response body in the `res.end()`

As another caveat, if `send` or a `return` value is not used, then the logger won't be able to pick up on the response body at all unless I am missing something and getting the data from the `ServerResponse` is possible.

I was unable to find a way to align the time text to the right side. If anyone else has an idea of how to do that with ANSI, please let me know.

Happy to chat more on the zeit slack about this as well!

EDIT: Updated screenshot

screen shot 2017-01-08 at 12 20 35 am

rauchg:
that looks really amazing! One way we could capture the response I assume is by patching the internal write method of the response object? The one that both `res.end` and `res.write` call

rickharrison:
Yea, we definitely could do that. I can take a stab and patching those methods inside of the `logRequest` method.

rauchg:
Also, I have a slightly different take on this now:

```jsx
{
  "dependencies": {
     "micro": "latest"
  },
  "devDependencies": {
     "micro-dev": "latest"
  }
}
```

- `micro-dev` would have `micro` in `peerDependencies`, rather than include it. 
- It works with the exact same signature as `micro`, but…
- We can add more options related to logging, for example whether to include headers, if you want to log only requests or only responses, etc.
- We get rid of `NODE_ENV` sniffing. If you run `micro-dev`, it logs. 
- Works really nicely for `package.json`. `start` can run `micro`, `dev` can run `micro-dev`

This way, `micro` remains lean both in download time, LOC/maintainability & performance!

rauchg:
https://github.com/zeit/micro-dev

onbjerg:
@rauchg The repository is empty -- forgot to push or do you need a PR? 😊

timneutkens:
I'm working on it @onbjerg 😄

onbjerg:
@timneutkens I couldn't wait so I implemented it as a simple wrapper function instead of a CLI tool: https://github.com/onbjerg/micro-visualize

timneutkens:
@onbjerg micro-dev is most likely coming today :)

onbjerg:
@timneutkens 😨

iamstarkov:
@timneutkens https://github.com/zeit/micro-dev is 404

onbjerg:
https://github.com/onbjerg/micro-visualize is still an option

PR #113, opened 2016-12-19
Added didBoot(port, host)

adnanbrq:
# Description
*fn:* didBoot (port, host)

micro now possesses a function called didBoot which will run after the server has booten up and can optionally be changed by the user when exporting a function with the same name.
By default it will just print the current known startup message. 

**Example**
`exports.didBoot = (port, host) => {
     // after boot
}`

### DIFF
--- a/bin/micro
+++ b/bin/micro
@@ -60,11 +60,15 @@ if ('/' !== file[0]) {
 require('async-to-gen/register')
 
 let mod
+let didBoot = function () {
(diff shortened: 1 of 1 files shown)


rauchg:
If you want custom execution hooks, I recommend you use the programmatic API, instead of invoking `micro`. We export a function [for this](https://github.com/zeit/micro#micro)

Thanks for the PR though!

PR #116, opened 2016-12-27
Proposal: Add 'text' method for body parsing

jmorrell:
While using micro to build out a GraphQL endpoint, I immediately ran into the issue of needing to get the graphQL query from the request object. I could use `raw-body` to do this directly, but seeing as how the `json()` function already has all of this logic, it seemed like a reasonable extension of the micro API.

I also understand that keeping the API tiny and simple is a priority for the project, so if this doesn't fit your direction, that's 👍

The included diff is just a quick proposal of the change. If this looks good I'll flesh out the tests and docs.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,6 +11,7 @@ const DEV = 'development' === process.env.NODE_ENV
 module.exports = exports = serve
 
 exports.run = run
+exports.text = text
 exports.json = json
 exports.send = send
 exports.sendError = sendError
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -39,18 +40,12 @@ async function run(req, res, fn, onError) {
   }
 }
 
-async function json(req, {limit = '1mb'} = {}) {
+async function text(req, {limit = '1mb'} = {}) {
```
jmorrell, line 43: Maybe `text` isn't the right name, maybe `rawBody`?

timneutkens:
@jmorrell Hi, could you take a look at https://www.github.com/timneutkens/micro-graphql? It might suit your needs already :)

jmorrell:
@timneutkens indeed that helps simplify things :)

timneutkens:
@jmorrell great 👍 the solution to just let `express-graphql` handle everything is pretty neat + it's maintained 😄

On the PR topic: I'm inclined to say this could be a `micro-rawbody` package. We've pretty much kept the core to a bare minimum. Things like urlencoded body parsing is also externalised.

jmorrell:
Makes sense 👍 Thanks for the feedback / help 🙂

timneutkens:
We're working on adding examples to the readme. Will try to add the graphql one too.

timneutkens:
@jmorrell FYI I just published https://www.npmjs.com/package/raw-text.

jmorrell:
@timneutkens nice 🙌

PR #118, opened 2017-01-03
Add method to parse plain text from request body

samtgarson:
This PR abstracts the logic to parse the raw buffer from a request so it can be used by `json` and also a new `text` method enabling parsing of plain text from request.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,6 +12,7 @@ module.exports = exports = serve
 
 exports.run = run
 exports.json = json
+exports.text = text
 exports.send = send
 exports.sendError = sendError
 exports.createError = createError
@@ -39,24 +40,38 @@ async function run(req, res, fn, onError) {
   }
 }
 
-async function json(req, {limit = '1mb'} = {}) {
+async function buffer(req, {limit = '1mb'} = {}) {
   try {
     const type = req.headers['content-type']
     const length = req.headers['content-length']
     const encoding = typer.parse(type).parameters.charset
-    const str = await getRawBody(req, {limit, length, encoding})
+    const buff = await getRawBody(req, {limit, length, encoding})
 
-    try {
-      return JSON.parse(str)
-    } catch (err) {
-      throw createError(400, 'Invalid JSON', err)
-    }
+    return buff
   } catch (err) {
     if ('entity.too.large' === err.type) {
       throw createError(413, `Body exceeded ${limit} limit`, err)
     } else {
       throw createError(400, 'Invalid body', err)
     }
+  }
+}
+
+async function json(req, options = {}) {
+  const str = await buffer(req, options)
+  try {
+    return JSON.parse(str)
+  } catch (err) {
+    throw createError(400, 'Invalid JSON', err)
+  }
+}
+
+async function text(req, options = {}) {
+  const str = await buffer(req, options)
+  try {
+    return str.toString()
+  } catch (err) {
+    throw createError(400, 'Invalid body', err)
   }
 }
 
--- a/README.md
+++ b/README.md
@@ -111,6 +111,26 @@ $ npm start
   }
   ```
 
+#### text
+
+**`text(req, { limit = '1mb' })`**
+
+- Use `require('micro').text`.
+- Buffers and parses the incoming body and returns it as plain text.
+- Exposes an `async` function that can be run with  `await`.
+- `limit` is how much data is aggregated before parsing at max. Otherwise, an `Error` is thrown with `statusCode` set to `413` (see [Error Handling](#error-handling)). It can be a `Number` of bytes or [a string](https://www.npmjs.com/package/bytes) like `'1mb'`.
+- If String parsing fails, an `Error` is thrown with `statusCode` set to `400` (see [Error Handling](#error-handling))
+- Example:
+
+  ```js
+  const { json, send } = require('micro');
+  module.exports = async function (req, res) {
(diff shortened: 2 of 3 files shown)


timneutkens:
Duplicate of #116. @leo what's your take on this?

samtgarson:
My bad, didn't see that one. I'm testing micro with a really tiny node server which is rendering JS components server side for a Rails app, so this would be really handy. Would be interested to know  your thoughts on this.

(I'm moving from express with `bodyParser.text()`)

timneutkens:
Personally I don't mind having it in there. Though most people will use either `json` or another format parser, for example: I created an [urlencoded-body-parser](https://github.com/timneutkens/urlencoded-body-parser) for use with micro.

samtgarson:
For sure—I figured it was so similar to `json` it would be a useful helper and easier to include here.

timneutkens:
@leo @rauchg could you take a look at this one?

lib/index.js
```diff
@@ -39,18 +40,14 @@ async function run(req, res, fn, onError) {
...
     const encoding = typer.parse(type).parameters.charset
-    const str = await getRawBody(req, {limit, length, encoding})
+    const buff = await getRawBody(req, {limit, length, encoding})
 
-    try {
-      return JSON.parse(str)
-    } catch (err) {
-      throw createError(400, 'Invalid JSON', err)
-    }
+    return buff
```
timneutkens, line 50: you  can `return await getRawBody(req, {limit, length, encoding})`

rauchg:
This makes a lot of sense to me

rauchg:
The question is, how common is this, that it merits having it in core?

timneutkens:
Just published https://www.npmjs.com/package/raw-text

rauchg:
I'd use `raw-text` from @timneutkens. Maybe feature it on the README ?

timneutkens:
@rauchg I will add an example to the new Readme 👍

samtgarson:
👌

dimapaloskin:
@timneutkens @rauchg @samtgarson 
`raw-text`  module is ok for most cases but there is one problem.

Look the code:
```js
const { json } = require('micro');
const parse = require('raw-text');

module.exports = async (req, res) => {
  const textBody = await parse(req);
  const jsonBody = await json(req);
  return { hello: 'world' };
};
```
json(req) will never be fulfilled because `raw-text` doesn't cache final buffer in the similar place with `json`.

Look this two lines please:
https://github.com/zeit/micro/blob/master/lib/server.js#L63
https://github.com/timneutkens/raw-text/blob/master/index.js#L7

`json` method keep buffer into map so we can use it as many times as we please.
`raw-text` keep it right in request (req.rawBody) and we can use it several times too.

But someone will have a problems in case if he want to use **both** this methods.

So I see two solutions:
1) Create `text` method and expose it like `json`.
2) Rewrite current current `json` code and keep raw body in the request.

What do you think about that? This is make sense?

samtgarson:
I would suggest:

```js
const parse = require('raw-text');

module.exports = async (req, res) => {
  const textBody = await parse(req);
  const jsonBody = JSON.parse(textBody);
  return { hello: 'world' };
};
```

dimapaloskin:
@samtgarson Yes, I understand how I can get the JSON from raw text. The main problem is caching. Let's think that someone need get raw body twice. Or just someone did grab all stream data before `json` calling. There is no any chance get the body again :(

In my code I need to use getRawBody before as "req" object get inside micro-service handler function. So inside this function I can't use `json` or `raw-text` or something else thats works with request stream. Because stream was read already. 

Will be very useful to make a ability to works with cache (or just raw body directly).

samtgarson:
I'm going to delegate to @timneutkens because I'm no expert (and I also don't understand why you'd need to use both methods in the same request... 😕)

dimapaloskin:
@samtgarson 
Something like that:

```js
const { run }  = require('micro');
const getRawBody = require('raw-body');

const file = 'path/to/other/micro.js';

module.exports = async (req, res) => {
  const body = await getRawBody(req);
  // there is some body transformation stuff
  const mod = require(resolve(file));
  run(req, res, mod);
}
```

dimapaloskin:
@samtgarson 
> I'm going to delegate to @timneutkens because I'm no expert

No problem. Thank you in any case!

timneutkens:
Okay, so the case here is that initially our solution to the multiple calls issues was adding `req.rawBody`. Later rauchg removed this and added the map. I wonder if we could just export the rawBodyMap for external modules to depend upon. That would certainly solve this issue.

dimapaloskin:
@timneutkens I don't think that rawBodyMap exporting is a good solution. At least it is looks a bit strange :) We can implement something like that:
```js
exports.getCachedBody = (req) => {
  return rawBodyMap.get(req);
}
```
But it is looks strange too because `.text` method is more logical approach.

Also I interesting the reason why rauchg change `req.rawBody` to map. Maybe it's can be reverted if it does not carry any critical fixes or improvements.

timneutkens:
> Also I interesting the reason why rauchg change req.rawBody to map. Maybe it's can be reverted if it does not carry any critical fixes or improvements.

this is because we don't want to add stuff to the req variable. Me might have to add the text method.
Lets open this ticket for now. cc @leo @rauchg

timneutkens:
Closing this in favor of a new issues I'm going to create right now.

PR #120, opened 2017-01-04
support PORT environment variable

cheslip:
This enables the use of a `PORT` environment variable that is commonly used with PaaS services such as Heroku and Cloud Foundry. Still defaults to `3000` if the env var is not set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ To run the microservice on port `3000` and localhost instead of listening on eve
 $ micro -p 3000 -H localhost sleep.js
 ```
 
+`micro` also supports using the `PORT` environment variable to set the port:
+
+```bash
+$ PORT=3000 micro sleep.js
+```
+
 ## Usage
 
 **Note**: `micro` requires Node `6.0.0` or later
--- a/bin/micro
+++ b/bin/micro
@@ -19,7 +19,7 @@ const args = parse(process.argv, {
   boolean: ['h'],
   default: {
     H: '0.0.0.0',
-    p: 3000
+    p: process.env.PORT || 3000
   }
 })
 let [,, file] = args._;


albinekb:
Add it to the readme too? @cheslip 🎉

albinekb (approved):

README.md
```diff
@@ -43,6 +43,12 @@ To run the microservice on port `3000` and localhost instead of listening on eve
 $ micro -p 3000 -H localhost sleep.js
 ```
 
+`micro` also supports using the PORT environment variable to set the port:
```
timneutkens: I would turn PORT into `PORT` (wrap it in '`')

timneutkens:
@leo @rauchg This one can be merged after the proposed change. Will this have any negative effect? I can imagine people having `PORT` defined and using `micro index.js`. Though that might be an edge case 🤔

timneutkens (commented):
Looks good 👍

timneutkens (approved):

rauchg:
Why not `micro -p $PORT`? Not being explicit about it is somewhat strange

cheslip:
If done that way it won't default back to `3000` if `$PORT` isn't set which is useful for testing locally. On the other hand you could just do `PORT=<port> micro -p $PORT` so just a matter of preference.

timneutkens:
@rauchg @cheslip Ah, I thought it was related to Heroku and Cloud Foundry not being able to do `-p` 😅 `micro -p $PORT` avoids weird bugs 👍

rauchg:
@cheslip in that case we should make sure we default to `3000` if `-p` is supplied with no arguments or empty ones?

timneutkens:
@rauchg I'll have a look into that 👍

cheslip:
yes that should achieve similar behavior, thanks for looking into it

Issue #121, opened 2017-01-06
Default to port 3000 when providing empty -p


timneutkens:
Issue from #120

leo:
Looks like [an issue](https://github.com/zeit/micro/blob/master/bin/micro#L22) in minimist...

timneutkens:
https://github.com/substack/minimist/issues/98 seems like it's not supported out of the box.

timneutkens:
@leo should we port to https://github.com/leo/args? 😏

timneutkens:
Ah it's based on minimist, nvm then 😉

timneutkens:
Documented in https://github.com/zeit/micro/pull/115

PR #123, opened 2017-01-07
Fallback to port 3000 when -p is empty

timneutkens:
#121

### DIFF
--- a/bin/micro
+++ b/bin/micro
@@ -81,7 +81,7 @@ if ('function' !== typeof mod) {
 
 let { port, host } = args
 
-if(typeof port === 'boolean') {
+if (typeof port === 'boolean') {
   port = 3000
 }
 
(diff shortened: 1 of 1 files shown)


timneutkens:
@leo can you review? Minimist will parse the following correctly:
`micro -p '' index.js`
`micro index.js -p ''`
`micro index.js -p`

Only case where it breaks is
`micro -p index.js`

since it will assign `index.js` to `-p`

bin/micro
```diff
@@ -79,7 +79,12 @@ if ('function' !== typeof mod) {
   process.exit(1)
 }
 
-const { port, host } = args
+let { port, host } = args
+
+if (typeof port === 'boolean') {
```
rauchg, line 85: maybe some comments here would be nice, to explain that the minimist behavior when no args are specified is to set it to `true`. Otherwise it's kinda non-obvious what's going on here

leo (approved):

bin/micro
timneutkens: Added 👍

PR #124, opened 2017-01-09
Don't log errors when testing

mxstbr:
Jest, the test runner, catches all console logs that happen during tests and outputs them at the end

I want to test that an error is thrown at a certain point in time, but because `micro` automatically `console.error`s errors it looks like the test failed, even though it didn't:

screen shot 2017-01-09 at 17 48 31

See that red stack trace? It's totally unnecessary and doesn't have to be there, as the test passed and this is expected behaviour.

> **Note:** This might be the wrong solution for the problem, throwing this PR out to kick off a discussion.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -114,7 +114,9 @@ function sendError(req, res, {statusCode, message, stack}) {
     send(res, 500, DEV ? stack : 'Internal Server Error')
   }
 
-  if (!TESTING) console.error(stack)
+  if (!TESTING) {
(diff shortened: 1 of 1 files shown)


timneutkens:
@mxstbr seems fair. Please have a look at https://travis-ci.org/zeit/micro/builds/190314237#L228

@rauchg what do you think?

timneutkens (approved):

timneutkens:
Approved 🙌

PR #129, opened 2017-01-22
Allow ``json`` to be called multiple times

onbjerg:
Consider a wrapper function that calls the ``json`` function, while wrapping another function that calls the ``json`` function as well, e.g.

```js
const { json } = require('micro')

const wrapper = function (fn) {
  return async function (req, res) {
    console.log('wrapper', await json(req))
    return await fn(req, res)
  }
}

module.exports = wrapper(async function (req, res) {
  console.log('main', await json(req))
  return 'stuff'
})
```

The call to ``json`` by the wrapper would block the next call indefinitely, since the stream from ``req`` has already been read.

This PR solves that.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -40,17 +40,20 @@ async function run(req, res, fn, onError) {
   }
 }
 
-const requestMap = new WeakMap()
+// maps requests to buffered raw bodies so that
+// multiple calls to `json` work as expected
+const rawBodyMap = new WeakMap()
+
 async function json(req, {limit = '1mb'} = {}) {
   try {
     const type = req.headers['content-type']
     const length = req.headers['content-length']
     const encoding = typer.parse(type).parameters.charset
 
-    let str = requestMap.get(req)
+    let str = rawBodyMap.get(req)
     if (!str) {
       str = getRawBody(req, {limit, length, encoding})
-      requestMap.set(req, str)
+      rawBodyMap.set(req, str)
     }
 
     try {
--- a/lib/index.js
+++ b/lib/index.js
@@ -40,14 +40,18 @@ async function run(req, res, fn, onError) {
   }
 }
 
+const requestMap = new WeakMap()
 async function json(req, {limit = '1mb'} = {}) {
   try {
     const type = req.headers['content-type']
     const length = req.headers['content-length']
     const encoding = typer.parse(type).parameters.charset
 
-    req.rawBody = req.rawBody || getRawBody(req, {limit, length, encoding})
-    const str = await req.rawBody
+    let str = requestMap.get(req)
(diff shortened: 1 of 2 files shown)


timneutkens:
@rauchg I monkey patched this in micro-dev. What would you prefer? If it's in micro core, then please do merge 😄

onbjerg:
@timneutkens It shouldn't be a monkeypatch since it's true for all wrapper functions that need to use ``json``.

lib/index.js
```diff
@@ -45,7 +45,9 @@ async function json(req, {limit = '1mb'} = {}) {
     const type = req.headers['content-type']
     const length = req.headers['content-length']
     const encoding = typer.parse(type).parameters.charset
-    const str = await getRawBody(req, {limit, length, encoding})
+
+    req.rawBody = req.rawBody || getRawBody(req, {limit, length, encoding})
```
rauchg: Maybe we should use a `WeakMap` so that we don't touch `req`?

rauchg:
This is a pretty big semantic change. We would have to document it.

lib/index.js
onbjerg: Where would you store the ``WeakMap``? Just outside the function?

onbjerg:
Sure, I'll document it, too.

lib/index.js
rauchg: Yep

rauchg:
Made a little clarification in the code. Let me know what you think

onbjerg:
Yeah, I think it makes sense to add a comment to clarify what's happening. I'm fixing the test.

onbjerg:
I documented the behavior. Any thoughts?

timneutkens (approved):
Looks good 👍

timneutkens:
Thing I'm wondering, should we add a `rawBody` or `body` function now. To expose the cache to external modules. We've been holding off on this since it would be possible without (https://github.com/timneutkens/raw-text).

onbjerg:
@timneutkens You can read the raw body with other modules from the request, since bodies can be formatted in different ways and thus are parsed differently. I don't think having a ``body`` function would serve any purpose

But maybe I'm wrong

timneutkens:
Yeah it's just the case of calling `json` on it and then `urlencoded` on it for example, where it would fail because the body was already parsed. It's an edge case though, I don't mind not supporting it.

onbjerg:
@timneutkens Is it even possible to have multiple bodies in the same request? I don't think I'm following

timneutkens:
I was explaining a pretty far fetched use case where you'd want to do something with the request body if json parsing fails. But nvm that, it's not important 👍

PR #131, opened 2017-01-23
Fix test example

BorisChumichev:
`test-listen` waits for server instance to be passed in as an argument.


timneutkens:
Good catch! Though, could you write the test example without `import` since all examples are using `require`.

BorisChumichev:
@timneutkens ✨

timneutkens (approved):

timneutkens:
Thanks! ❤️

PR #135, opened 2017-01-30
Remove greenkeeper email option

timneutkens:
Saw it was removed from serve because it didn't do anything. Therefor removed here.


leo:
Thank you!

PR #136, opened 2017-01-31
Return 204 No Content if value is null

yagop:
I don't now if this wold be useful but here it is.

Example:
```js
module.exports = async function () {
  return null
}
```
```bash
➜ http localhost:3000
HTTP/1.1 204 No Content
Connection: keep-alive
Date: Tue, 31 Jan 2017 12:37:26 GMT
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -31,8 +31,12 @@ async function run(req, res, fn, onError) {
   try {
     const val = await fn(req, res)
 
-    // return a non-null value -> send
-    if (null !== val && undefined !== val) {
+    // return 204 No Content if value is null
+    if (null === val) {
+      send(res, 204, null)
+    }
+    // return a undefined-null value -> send
+    if (undefined !== val) {
       send(res, res.statusCode || 200, val)
     }
   } catch (err) {


yagop:
I did the PR because I couldn't find another way to end the request rather than:

```js
module.exports = async function () {
  // ...
  return '' // Or send(res, 200, null)
}
```

timneutkens (commented):
You can actually do this by creating a wrapping function, haven't tried this snippet. But it should get you started ❤️ 

Edit: Just tested. Here's the functioning wrapper:

```js
const { send } = require('micro')

function wrapNull(fn) {
  return async (req, res) => {
    const val = await fn(req, res)
    if(null === val) {
      send(res, 204, null)
      return
    }

    if (undefined !== val) {
      return val
    }
  }
}

module.exports = wrapNull(async function (req,res) {
  return null
})
```

timneutkens:
cc @leo @rauchg Do you want to merge this in or not? It's definitely possible by wrapping.

leo:
[Done!](https://github.com/zeit/micro/commit/65dbed8889c5634a237d74ff5e8a6746df76c63b)

PR #141, opened 2017-02-04
Log error stack only in development

fridays:


jamo:
Not sure of swallowing errors in prod is such a great idea... (Only viewed this on mobile so I'll revise it later today)

timneutkens:
I'm not sure why you wouldn't want them on production. Can you elaborate @fridays?

You could do this to not log them:

```js
const {send} = require('micro')

function sendError(req, res, {statusCode, message, stack}) {
  if (statusCode) {
    send(res, statusCode, DEV ? stack : message)
  } else {
    send(res, 500, DEV ? stack : 'Internal Server Error')
  }
}

function wrapError (fn) {
  return async (req, res) => {
    try {
      await fn(req, res)
    } catch (err) {
      sendError(req, res, err)
    }
  }
}

async function handler (req, res) {
  throw new Error('Random error occured')
}

module.exports = wrapError(handler)
```

fridays:
The documentation says

> Important: during development mode (if the env variable NODE_ENV is 'development'), error stacks will be printed as console.error and included in the responses.

But the code does `console.error(stack)` in any case unless `NODE_ENV` is `test`. Maybe I am missing something. Should it print a stack when I do `throw createError(404, 'Not found')` in a request?

timneutkens:
@rauchg What would you prefer? 
- Fix the documentation
- Fix the behaviour to match the documentation

PR #149, opened 2017-02-07
Allow async-to-gen register to be loaded more than once

timneutkens:
@leo as we talked about.

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -4,7 +4,6 @@
 const path = require('path')
 
 // Packages
-const asyncToGen = require('async-to-gen/register')
 const updateNotifier = require('update-notifier')
 const nodeVersion = require('node-version')
 const args = require('args')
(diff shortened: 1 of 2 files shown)


timneutkens:
@leebyron 
Hi Lee, `async-to-gen/register` registers the require hook at the moment you call `require('async-to-gen/register')` instead of when calling the exported function. Because of this you can't use it twice with different configs. Is there any reasoning behind this we haven't thought of?

leebyron:
The require hook has a global effect on the environment, so a second call with a different configuration would overwrite the first.

Since the require hook is a development-time aid for what will ultimately be the configuration to the transform script, I thought it would be too confusing (or a possible "foot gun") to suggest the config be overwritten at runtime

timneutkens:
@leebyron yeah it would introduce some unwanted effects when used improperly, so I guess it's best for most use cases 😄

leo:
Fixed! Thanks guys 😊

PR #150, opened 2017-02-08
Include lib and bin into async-to-gen compilation

diorahman:
It seems the current regex rejects `./lib` and `./bin`. But allow `micr/lib` and `micr/bin` instead.

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -60,7 +60,7 @@ if (!isAsyncSupported()) {
   const directoryName = path.parse(path.join(__dirname, '..')).base
 
   asyncToGen({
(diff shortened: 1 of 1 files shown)


timneutkens:
Async to gen matches that regex against absolute paths, so it Should be fine How it is right now. #149 solves a bug in the registering, after that all Should be fine

timneutkens:
Currently on my phone though. Will check it out later 🙌🏻

diorahman:
OK. But it seems micro is getting fatter than before hehe

leo:
https://github.com/zeit/micro/pull/150#issuecomment-278261744 😊  Thanks for your effort though, @diorahman!

PR #152, opened 2017-02-11
Include whole entrypoint directory to be compiled

timneutkens:
@leo this was in my previous PR to fix compilation. Should still be included.


leo:
Yep, I also still remember why we need that. Thanks! ❤️

timneutkens:
👍

PR #153, opened 2017-02-11
Use bluebird instead of Q

timneutkens:


leo:
👍

But please remove the `const {coroutine} = Promise` and use `Promise.` for these instances... 😊

PR #154, opened 2017-02-12
[WIP] Convert to Bluebird

lukeed:
_This isn't ready for merging yet!_

### Why

For a few months now, I've been using my private fork of `micro`, where I converted it to be powered by Bluebird. I maintained the fork because I was under the impression that a core goal of `micro` was to stick to the native platform. 

I recently released my fork as `polkadot`, only to realize that `micro` migrated to the `Q` library. 😲 To me, this signaled that a Promise-library was welcomed, so I could actually kill my fork & apply the Bluebird changes directly to `micro`. After talking to @leo and @timneutkens, the idea was welcomed! 😀 

Here are the [performance gains](https://github.com/lukeed/polkadot/blob/master/benchmarks/readme.md) I saw with `polkadot` compared to the latest version of `micro`. Crazy!

### Currently...

- The server & its promise-based methods are fully powered by Bluebird's coroutine. 

    _The `wrk` benchmark (see above) increased from 19k req/sec to 46k req/sec._

- In addition, CLI usage dropped the conditional/any usage of `async-to-gen`. This was the biggest hindrance to `micro`'s performance.
    
  _This hurt execution speed **AND** memory efficiency. Removing this drops [memory usage](https://github.com/lukeed/polkadot/blob/master/benchmarks/readme.md#memory-usage) from ~70 MB to ~20 MB._

- ~~The main module (`lib/index.js`) had a 3rd argument that always relied on its default value. Removed the parameter and defined a scoped constant instead.~~

- Rewrote tests from `async/await` functions to `*gen/yield` functions. ~~**This needs to be improved**.~~ AVA's internal copy of Babel was compiling these; should use raw generators.

### To Do

- [x] Improve/unlock all tests
- [x] Re-enable `async` function exports.

   _At server start-up, the user-defined module will be transpiled (without an ongoing runtime!) into a `GeneratorFunction` & then passed to Bluebird. This is what Fly does with `fly-esnext` and there's **hardly** any overhead._

  _I'm currently writing a module that mimics `fly-esnext`, but for non-Fly/universal use._

- [x] **Allow for synchronous module functions?** The tests define that a user can export a generic function; but is this still desired? I can add a simple wrapper if so, but want to double-check first.


diorahman:
@lukeed how do you measure the memory usage? Thanks!

lukeed:
@diorahman I just have activity monitor (OSX) open & watch the `node` process that powers the server after 9,000 requests. I'm running the `ab` script (defined in my link), so the testing conditions are controlled. The environment is consistent for back-to-back tests.

kandros:
Looks like it has been already implemented https://github.com/zeit/micro/commit/376120e95a7ffd490a39bcf088a75129a1fbff97

lukeed:
@kandros Nope, incomplete 😄 This is a continuation of what was done.

lib/index.js
```diff
@@ -1,13 +1,15 @@
 // Packages
 const detect = require('detect-port')
+const {coroutine} = require('bluebird')
 
 // Ours
 const serve = require('./server')
 const listening = require('./listening')
 const getModule = require('./module')
 
-module.exports = (file, flags, module = getModule(file)) => {
```
timneutkens, line 9: @lukeed this was done to facilitate micro-dev, which is not released yet.

lukeed: ah, ok. will revert

timneutkens: ❤️ Basically what we do in micro-dev is require the file directly and then pass a different module 😄

lukeed: 👍 Updated original PR text.

lukeed:
My before/after tests (using the same `wrk` bench) show that `micro` jumps from 14-15k req/sec (`7.0.6`) to 45-47k req/sec (this PR).

iamstarkov:
My suggestion to this PR is not to use hard coded Promise implementation, but rather offer the way to use 3rd party A+ implementation and use native Promises by default.

floatdrop:
@lukeed did you run bench it agains native `async/await`? Looks like this PR removes `is-async-supported` and binds `mirco` to bluebirds coroutines. I think this contradicts with first [feature](https://github.com/zeit/micro#features):

> Designed for usage with async and await

Also `rewrite-async` will do funny things with variable names like `const asyncHelper`.

leo:
Now that native support for `async` and `await` has [landed](https://twitter.com/notquiteleo/status/834330621107433472), I'd rather wait until we have it in LTS and then remove on-the-fly-transpilation from the micro CLI (the programmatic API already isn't being transpiled at all).

Thank you for your help and inspiration @lukeed, I learned a lot from this... 😊

lukeed:
Hey @leo, you bet! I held off on finalizing the generator Regex rewrites because I figured you'd want to move in this direction.

However, I ran the 7.6 `async` benchmarks for `micro` 7.0.1. (This version had no `Q` & I gutted `async-to-gen`). The req/s (wrk bench) is 30k. This is still a big jump for micro, but it's only for those who are running with harmony or 7.6.

The pure Bluebird implementation is still running at 45k across all versions.

Issue #145, opened 2017-02-05
CLI transpiler doesn't run in Windows environment


chrisbeaver:
I've been playing with the code here trying to get the async/await calls firing which I understand use the is-async-supported, and async-to-gen libraries. As a basic example, I follow the instructions for installing the example provided: https://github.com/zeit/micro/tree/master/examples/external-api-call

Pulling the code in on mac, npm install, npm run start. This works as expected. 

Doing the identical thing in Git Bash on Win 10 results in a reference to the async function.

![micro](https://cloud.githubusercontent.com/assets/5077917/22629771/9c623d4e-ebba-11e6-8896-01709426fd95.png)

This code is not being transpiled. It may be an issue with that specific library which maybe I should refer this issue to that repo.

timneutkens:
Hmm good catch, I don't have a windows machine around to test. @leo could you check this out? ❤️ 

Regarding your last question. I'm pretty sure it's an issue on our side, since it has worked before 😅

leo:
It's an issue with the `file` variable [here](https://github.com/zeit/micro/blob/master/bin/micro.js#L63). It's somehow not including it in the transpilation...

chrisbeaver:
I'm wondering if it's being passed a symbolic link. That is the issue most of the time I find with node modules.

edit: It's not passed a symbolic link. It's a windows absolute path: 
C:\Users\Chris\AppData ... \index.js

zonak:
The issue is the regular expression used to determine which files to transpile.

The path of the filename in Windows contains backslashes which, if not escaped, will cause a problem with the regular expression and therefore not match the `index.js` file for transpilation.

A quick workaround would be to escape the special characters in the file path:

```javascript
asyncToGen({
  includes: new RegExp(`.*${directoryName}?${pathSep}(lib|bin)|${file.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&')}.*`),
  excludes: null,
  sourceMaps: false
})
```

which we can also make conditional if we want to avoid doing it when not on Windows.

Also, was wondering if there is a specific reason why instead of using `path.sep` the path separator is defined as:

```javascript
const pathSep = process.platform === 'win32' ? '\\\\' : '/'
```

fip3torrejon:
I followed instructions for https://github.com/zeit/micro/tree/master/examples/external-api-call on GitBash for Windows 7, and Ubuntu 16.04.1 terminal... got the same error both.
[...]

PR #158, opened 2017-02-14
Fix transpilation on Windows platform

manrueda:
This change fix the issue #145, where the code was not transpiled on Windows platform.

The issue was a escaping issue with the `\` directory separator of windows in the regex that check if the file exists or not.


manrueda:
Close this because an issue in the commit, wrong data. I will send a new one... sorry

PR #163, opened 2017-02-23
Remove bluebird.coroutine

floatdrop:
~~Since Node.JS 7.6.0 supports `async`/`await` without flags.~~

Coroutine is used just for couple `yields` in code and can be safely replaced with Promise chain.

Benchmark results (based on [`polkadot`](https://github.com/lukeed/polkadot/tree/master/benchmarks) benchmarks). 

Commands for benchmarking:

- **wrk**: `wrk -t12 -c400 -d30s http://localhost:3000/`
- **ab**: `ab -c4 -n5000 http://0.0.0.0:3000/`

And numbers:

&nbsp; | Node 7.5.0 (wrk) | Node 7.5.0 (ab) | Node 7.6.0 (wrk) | Node 7.6.0 (ab)
------- | ---------- | ------------- | ------------ | -------------
7.0.6 | 6553.51 | 1990.89  | 7854.89  | 2609.53 
master | 12353.73 | 3036.74  | 15519.98  | 4412.05 
PR | 14492.10 | 3698.73 | 17533.89 | 5631.77

On `7.5.0` app was launched with `micro`, on `7.6.0` app was launched with `node`.

Performance can be doubled with [uWebSockets](https://github.com/uWebSockets/uWebSockets) (up to 31843 rps).

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -5,6 +5,9 @@ const server = require('http').Server
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
 const isStream = require('isstream')
+const Promise = require('bluebird')
+
+const {resolve} = Promise
 
 const DEV = process.env.NODE_ENV === 'development'
 const TESTING = process.env.NODE_ENV === 'test'
@@ -18,7 +21,7 @@ exports.sendError = sendError
 exports.createError = createError
 
 exports.run = (req, res, fn) =>
-  Promise.resolve(fn(req, res))
+  resolve(fn(req, res))
     .then(val => {
       if (val === null) {
         send(res, 204, null)
--- a/lib/server.js
+++ b/lib/server.js
@@ -5,16 +5,11 @@ const server = require('http').Server
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
 const isStream = require('isstream')
-const Promise = require('bluebird')
-
-const {coroutine, resolve} = Promise
 
 const DEV = process.env.NODE_ENV === 'development'
 const TESTING = process.env.NODE_ENV === 'test'
 
-const serve = fn => server(coroutine(function * (req, res) {
-  yield exports.run(req, res, fn)
-}))
+const serve = fn => server((req, res) => exports.run(req, res, fn))
 
 module.exports = exports = serve
 
@@ -22,22 +17,19 @@ exports.send = send
 exports.sendError = sendError
 exports.createError = createError
 
-exports.run = coroutine(function * (req, res, fn) {
-  try {
-    const val = yield resolve(fn(req, res))
+exports.run = (req, res, fn) =>
+  Promise.resolve(fn(req, res))
+    .then(val => {
+      if (val === null) {
+        send(res, 204, null)
+      }
 
-    if (val === null) {
-      send(res, 204, null)
-    }
-
-    // Return a undefined-null value -> send
-    if (undefined !== val) {
-      send(res, res.statusCode || 200, val)
-    }
-  } catch (err) {
-    sendError(req, res, err)
-  }
-})
+      // Return a undefined-null value -> send
+      if (undefined !== val) {
+        send(res, res.statusCode || 200, val)
+      }
+    })
+    .catch(err => sendError(req, res, err))
 
 // maps requests to buffered raw bodies so that
 // multiple calls to `json` work as expected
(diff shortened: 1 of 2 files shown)


timneutkens:
Unfortunately we have to support LTS. So this can't be merged (yet).

timneutkens:
Do note `isAsyncSupported` will detect if you have native async/await and not load the async transform. So that will work immediately.

floatdrop:
@timneutkens Yeah, I know. This PR was alternative to #154 – but @leo already closed it.

> Do note isAsyncSupported will detect if you have native async/await and not load the async transform. So that will work immediately.

It will, but it can't undo [`coroutine`](https://github.com/zeit/micro/pull/163/files#diff-c945a46d13b34fcaff544d966cffcabaL25) wrapper.

iamstarkov:
@timneutkens, [LTS support](https://github.com/nodejs/lts) means that this wont be merged until 18 april 2018, am i right?

floatdrop:
In fact `bluebird` can just be dropped. ~~However this will break `micro` code, that rely on generators~~ Exporting generators to micro is currently not working.

floatdrop:
I ran same test as in [`polkadot`](https://github.com/lukeed/polkadot/tree/master/benchmarks). Results are in issue description.

lukeed:
I don't understand 😖 Your PR is to remove Bluebird, yet you include benchmarks that prove its the fastest, even with the most minimal implementation?

Remember, too, that most people don't actually run on `7.x`, or with any harmony flags on. So removing Bluebird would be further detrimental to all those use cases.

floatdrop:
@lukeed I've included benchmark with bluebird just to see, how much rps can be squeezed from `micro`.

`master` is using bluebird, but it still slower, than `use-async`, which does not.

I'm not proving, that `bluebird` is slow (it is fast as hell), but that `bluebird.coroutine` can be safely removed with improved performance. Benchmarks on `7.5.0` were ran without harmony flags, so I can anticipate same performance increase on `6.x` versions of `Node.JS`.

Users can always do `global.Promise = require('bluebird');` in their code to boost up performance (or fallback on different implementation).

lukeed:
Ah, I see. I forgot that it was current `master`. 

Well, current `master`'s use of Bluebird is wrong/incomplete. We were talking on Slack while it was happening, and then it was decided I would take over (can see my PR).

The main culprits with current `master` are `serve()` and `run()`;  see: https://github.com/zeit/micro/pull/154/files#diff-c945a46d13b34fcaff544d966cffcaba. 

Currently, `serve` is wrapping a new generator only to yield an existing coroutine, when instead, the coroutine can just be returned because it will be yielded elsewhere. More work than it needs to do.
And `run`, is wrapping a `fn` inline on _every_ request, even though it's already a yieldable/resolvable sequence.

All in all, current `master` is "tripping over itself" and doing more work than it needs to in order to execute a single response.

Not that it matters at this point, because I agree with you re: 
> Users can always do global.Promise = require('bluebird'); in their code to boost up performance (or fallback on different implementation).

...but if you benchmark my PR (complete Bluebird integration), I'm sure you'll find different results. 😃

floatdrop:
@lukeed good to know, that there is more space for improvements! 😁

Benchmarking #154 gives me 15964.01 rps and 16982.42 rps on `7.5.0` and `7.6.0`.

Here some profiling data for this PR: https://gist.github.com/floatdrop/e1695d1d539af54c21ca44305320cade#file-profiling-txt-L847

Most heavy call from `micro` code  is `Runtime_InstanceOf` from `micro/lib/server.js:72:14` wich is `send` function. I don't think it will be easy to optimise.

lukeed:
Ooh, awesome! Thanks for that!

It looks like line `72` is actually the middle of `json()`. Did you make any customizations to the code you ran?

[Looking at it](https://github.com/lukeed/micro/blob/2311027bc65b8e67a47e4d13ac0f219c9431d031/lib/server.js#L56-L62), this makes perfect sense to me. Hoisting & wrapping with `resolve()` is not ideal, and that's why I left it as a [nested `try/catch` in polkadot](https://github.com/lukeed/polkadot/blob/master/index.js#L79-L83), which is also not great, but a better option. 

Perhaps your tooling can validate that?

floatdrop:
@lukeed this perf is for this PR, not for `polkadot` 😅

timneutkens (approved):
Looks good 👍

leo:
Thank you! 😊

PR #164, opened 2017-02-26
Add `status` fallback to error handling

dotcypress:
Add `status` fallback to error handling.
This change makes some useful libraries(http-assert) compatible with `micro` 🚀

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -126,9 +126,9 @@ function send(res, code, obj = null) {
   res.end(str)
 }
 
-function sendError(req, res, {statusCode, message, stack}) {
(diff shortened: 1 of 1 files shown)


timneutkens:
Alternative approach:

```js
function wrap (fn) {
  return async (req, res) => {
    try {
      const value = await fn(req, res)

      return value
    } catch (err) {
      if(err.status) {
        err.statusCode = err.status
      }

      throw err
    }
  }
}

const assert = require('http-assert')

module.exports = wrap(async () => {
  const test = 'abc'
  assert(test === 'abc', 404, 'Not found')

  return 'Ok'
})
```

leo:
👍 Thanks!

PR #168, opened 2017-03-09
Convert port to number

nkzawa:
This causes to display a wrong message like the following.

screen shot 2017-03-09 at 15 34 46


timneutkens (approved):
💯

PR #171, opened 2017-03-10
Allow numbers as responses.

knpwrs:
I was trying to write a microservice earlier today and came to realize that you can't return numbers from your handlers. Plain numbers are actually valid JSON, so I think it makes sense to support them.

Might I also suggest linting _after_ running tests? Sometimes it would be nice to run `npm test` and get some actual test output without having to worry about style problems which can be fixed later.


timneutkens (approved):
Seems reasonable since we also allow plain strings. 👍

PR #172, opened 2017-03-11
Export async function lib/index.js

floatdrop:
This makes code a little bit nicer. Async is used in this file already – so this should not be a problem.


timneutkens (approved):
👍

PR #178, opened 2017-03-13
Simplify returnJSON function

floatdrop:
Passing `resolve` and `reject` around can become messy and can make transition to `async`/`await` harder.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -5,9 +5,7 @@ const server = require('http').Server
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
 const isStream = require('isstream')
-const Promise = require('bluebird')
-
-const {resolve} = Promise
+const {resolve} = require('bluebird')
 
 const DEV = process.env.NODE_ENV === 'development'
 const TESTING = process.env.NODE_ENV === 'test'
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -38,31 +38,30 @@ exports.run = (req, res, fn) =>
...
-    resolve(JSON.parse(str))
+    return JSON.parse(str)
   } catch (err) {
-    reject(createError(400, 'Invalid JSON', err))
+    throw createError(400, 'Invalid JSON', err)
   }
 }
 
-exports.json = (req, {limit = '1mb'} = {}) => new Promise((resolve, reject) => {
+exports.json = (req, {limit = '1mb'} = {}) => resolve().then(() => {
```
albinekb, line 47: `Promise.resolve()`?

floatdrop: @albinekb it is already destructed [here](https://github.com/zeit/micro/blob/master/lib/server.js#L10). Actually `const Promise = require('bluebird')` can be replaced with `const {resolve} = require('bluebird')` now.

albinekb: Oh, didn't see 😄  not used to Promise being destructed.

albinekb (approved):

timneutkens (approved):
Neat 👍

PR #179, opened 2017-03-14
Null value from fn should cause one res.end call

floatdrop:
Small bug 🐛 

I wonder, why node is not panic on this situation.


timneutkens (approved):
👍 

> I wonder, why node is not panic on this situation.

It probably will on a larger response 😅

floatdrop:
> It probably will on a larger response

`null` can't get larger 😅

timneutkens:
@floatdrop True :joy: I meant if we did something other than sending `null` we would have noticed 😄

PR #180, opened 2017-03-14
Undefined from fn should not hang request

floatdrop:
Not-quite-small bug 🐞 

I'm not sure, this is the right way to deal with `undefined` from `fn`, but surely one of. Other way – to throw error.


timneutkens (approved):
I believe this will be ok for these specific cases. Especially since RN it would never end the request right? Sending a 204 on `undefined` seems reasonable.

timneutkens:
cc @leo

leo:
👍 Thanks!

PR #183, opened 2017-03-14
Fix for #164 – status should not be ignored

floatdrop:
Fallback `statusCode || status` was under `if (statusCode) { ... }` – so it never worked (because `statusCode` is always truthy.


timneutkens (approved):
👍

PR #189, opened 2017-03-16
Support HEAD method

### LINKED ISSUE #184: Support HEAD method
For now `micro` is not checking `req.method` in any way – it would be super-handy to drop body automatically on `HEAD` methods.

floatdrop:
This is one way to solve `HEAD` requests. I don't like fiddling with `err.mesage`, but I don't like idea to pass `req` into `send` method.

Closes #184

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -18,8 +18,10 @@ exports.send = send
 exports.sendError = sendError
 exports.createError = createError
 
-exports.run = (req, res, fn) =>
-  new Promise(resolve => resolve(fn(req, res)))
+exports.run = (req, res, fn) => {
+  res.request = req
+
+  return new Promise(resolve => resolve(fn(req, res)))
     .then(val => {
       if (val === null || val === undefined) {
         send(res, 204, null)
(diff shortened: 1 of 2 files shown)


timneutkens (approved):
👍

floatdrop:
@tim @leo another way is to add `req` to `res` object and move checks into `send` method.

rauchg:
I'm -1 on this. I think `send` would be too magical

floatdrop:
@rauchg yeah, I don't like this too, but is there a better way to support HEAD methods?

leo:
https://github.com/zeit/micro/pull/189#issuecomment-287581566

@rauchg please describe how [this](https://github.com/zeit/micro/pull/189#issuecomment-287621168) can be solved.

floatdrop:
@leo it is possible to wrap `send` and `sendError` in user land, guess something like module `micro-methods` will do.

ianstormtaylor:
Is this really not coming back? When would you ever want `HEAD` to include the body when using `send()`? I don't get how that's too magical... 

Edit: Turns out Node's core already handles `HEAD` internally to remove the body: https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L573

PR #190, opened 2017-03-16
💯 test coverage

floatdrop:
Some places can be better. For example testing for call to `console.error` is quite dirty. May be logging should be improved?

P.s. still `87.18%` for branches.


timneutkens (approved):
Nice!

Issue #174, opened 2017-03-11
Should we add a raw text method?


timneutkens:
See #118

leo:
Please do [this](https://github.com/zeit/micro/pull/118#issuecomment-270204610), if you have time.

floatdrop:
@leo can you elaborate on this? `.text` method will not introduce new dependencies (because `raw-body` is used for `.json` already) and will do nice decomposition of `.json` method. Seems logical to expose such method, if `mirco` already have this code inside.

Also I think #207 caused by confusing `.json` method with parsing plain bodies.

leo:
Resolved with https://github.com/zeit/micro/pull/208

PR #203, opened 2017-03-25
Support handling query strings

malexdev:
Handling query strings is simple, but it seemed like a good thing to add to the project.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -2,6 +2,7 @@
 const server = require('http').Server
 
 // Packages
+const url = require('url')
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
 const isStream = require('isstream')
@@ -66,6 +67,10 @@ exports.json = (req, {limit = '1mb'} = {}) => Promise.resolve().then(() => {
       throw createError(400, 'Invalid body', err)
     }
   })
+})
+
+exports.query = req => Promise.resolve().then(() => {
+  return url.parse(req.url, true).query
 })
 
 function send(res, code, obj = null) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -72,9 +72,7 @@ exports.json = (req, {limit = '1mb'} = {}) => Promise.resolve().then(() => {
   })
 })
 
-exports.query = req => Promise.resolve().then(() => {
-  return url.parse(req.url, true).query
-})
+exports.query = req => url.parse(req.url, true).query
 
 function send(res, code, obj = null) {
   res.statusCode = code
--- a/test/index.js
+++ b/test/index.js
@@ -15,29 +15,27 @@ const getUrl = fn => {
 }
 
 test('support query string', async t => {
-  const fn = async req => {
-    const q = await query(req)
-    return q.test
-  }
-
   const qs = {test: 'test'}
-  const url = await getUrl(fn)
-  const res = await request({url, qs})
 
-  t.deepEqual(res, 'test')
+  const fn = async req => {
+    const q = query(req)
+    t.deepEqual(q, qs)
+  }
+  
+  const url = await getUrl(fn)
+  await request({url, qs})
 })
 
 test('support query string with multiple options', async t => {
+  const qs = {a: 'a', b: 'b', c: 'c'}
+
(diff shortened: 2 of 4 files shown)


test/index.js
```diff
@@ -6,14 +6,40 @@ const resumer = require('resumer')
...
+  const fn = async req => {
+    const q = await query(req)
+    return q.test
+  }
+
+  const qs = {test: 'test'}
+  const url = await getUrl(fn)
+  const res = await request({url, qs})
+
+  t.deepEqual(res, 'test')
```
floatdrop: You can just call `t.deepEqual` in req handler.

test/index.js
```diff
@@ -6,14 +6,40 @@ const resumer = require('resumer')
...
+  const fn = async req => {
+    const q = await query(req)
+    return {a: q.a, b: q.b, c: q.c}
+  }
+
+  const qs = {a: 'a', b: 'b', c: 'c'}
+  const url = await getUrl(fn)
+  const res = await request({url, qs})
+
+  t.deepEqual(res, JSON.stringify({a: 'a', b: 'b', c: 'c'}))
```
floatdrop: Here too – `JSON.stringify` is not good approach.

lib/server.js
```diff
@@ -68,6 +69,10 @@ exports.json = (req, {limit = '1mb'} = {}) => Promise.resolve().then(() => {
   })
 })
 
+exports.query = req => Promise.resolve().then(() => {
```
floatdrop: There is no need for `Promise` here – no async work is done.

floatdrop (changes requested):

lib/server.js
malexdev: Gotcha. Thought we might want to keep it the same as the `.json` API so made it use the promise. Makes sense to make it sync though.

test/index.js
```diff
@@ -6,14 +6,38 @@ const resumer = require('resumer');
...
 
   return listen(srv);
 };
 
+test('support query string', async t => {
+  const qs = {test: 'test'}
+
+  const fn = async req => {
+    const q = query(req)
+    t.deepEqual(q, qs)
```
floatdrop: With latest fix (7.2.2) – you need to return here something, otherwise request will hang.

test/index.js
```diff
@@ -6,14 +6,38 @@ const resumer = require('resumer');
...
+  const url = await getUrl(fn)
+  await request({url, qs})
+})
+
+test('support query string with multiple options', async t => {
+  const qs = {a: 'a', b: 'b', c: 'c'}
+
+  const fn = async req => {
+    const q = query(req)
+    t.deepEqual(q, qs)
```
floatdrop: And here too.

floatdrop (changes requested):

floatdrop (commented):
`url.parse` is not quite get things right – maybe it is better to use [`query-string`](https://github.com/sindresorhus/query-string) module or look at [URL class](https://nodejs.org/api/url.html#url_the_whatwg_url_api). Anyways parsing query string is trivial task, should it have own helper?

malexdev:
If you don't think it would be a good fit for micro that's fine, I just thought it'd be helpful to have built in.  Also, what is it that `url.parse` doesn't get right? I haven't seen any issues with it so far.

Not trying to be combative, I'm open to whatever you guys want to do. If you don't want querystring parsing built into Micro that's fine, just reject the PR. If you prefer to use the `query-string` module, that's fine, I'll modify my PR.

floatdrop:
@malexdev standart module does not support arrays in query or nested objects for example:

```js
url.parse('http://google.com/?query=1&a[0]=1&a[1]=1', true).query
// { query: '1', 'a[0]': '1', 'a[1]': '1' }
```

I can not merge or reject PR, just judging by #174/#184 this kind of PR are preferred as readme sections or separate modules. It kinda tough way, but keeps code and package small and shiny.

leo:
Thank you for your effort, guys! But this is not something we'd like to see built-in. The point of `micro` is to only provide very little (the most important) abstractions for microservices.

PR #209, opened 2017-03-27
Added `.buffer` and `.text` helpers

floatdrop:
Based on #208.

- added `.buffer` and `.text` helpers
- fixed wrong error message `Invalid body` on JSON parsing


leo:
Resolved in https://github.com/zeit/micro/pull/210

PR #218, opened 2017-04-02
Added chat-app example

coffee-mug:
Added code from chat-app tutorial [Chat app tutorial](https://zeit.co/docs/examples/chat)

### DIFF
--- a/examples/socket.io-chat-app/index.js
+++ b/examples/socket.io-chat-app/index.js
@@ -1,0 +1,20 @@
+const micro = require('micro'),
+      fs = require('fs');
+
+const server = micro(async (req, res) => {
+  const html = fs.readFileSync(__dirname + '/index.html');
+
+  console.log("Serving index.html");
+
+  res.end(html);
+});
+
+const io = require('socket.io')(server);
+
+// socket-io handlers are in websocket-server.js
+require('./websocket-server.js')(io);
+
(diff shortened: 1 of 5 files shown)


leo:
Sweet, thanks!

examples/socket.io-chat-app/index.js
```diff
@@ -0,0 +1,20 @@
+const micro = require('micro'),
+      fs = require('fs');
+
+const server = micro(async (req, res) => {
+  const html = fs.readFileSync(__dirname + '/index.html');
```
albinekb, line 5: You don't want to read the index contents every time the route is hit, specially not *sync*, this will block other incoming connections

coffee-mug: Hum yes did not think of that, I used it naively to serve the html but you're right it will just create troubles. What would be the right way to do it with micro ?

leo: Lemme fix - great find, @albinekb

leo: Fixed in 66fcf94acdb4d2f73c22f62f05dd7217b81cf0e2 - sync is okay on startup

coffee-mug: Awesome ! Thanks a lot @albinekb @leo !

PR #219, opened 2017-04-02
Reduce package size by 1MB

timneutkens:
Moved all dx modules to micro-dev

old:
old

new:
new

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,3 @@
-// Packages
-const detect = require('detect-port');
-
 // Ours
 const serve = require('./server');
 const listening = require('./listening');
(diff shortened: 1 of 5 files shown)


floatdrop:
@timneutkens can you describe, how developer should now use `micro`? Also, why `args` is not dx module too?

lib/index.js
```diff
@@ -9,33 +6,28 @@ const getModule = require('./module');
 module.exports = async (file, flags, module = getModule(file)) => {
```
albinekb, line 6: this does not need to be `async` anymore since there's no `await` in it

floatdrop:
As side note – we can reduce package even more splitting `args`, `is-async-supported`, `node-version` and  `async-to-gen` to `micro-cli`.

timneutkens:
See #260

PR #226, opened 2017-04-08
Replace detect-port with get-port

floatdrop:
More lightweight and pretty module for getting available port.


albinekb (approved):

jamo:
I think the plan is to soon going to drop the logic which may randomly pick the port in production if the default one is taken. It may easily cause issues that are hard to debug.
cc/ @timneutkens

leo:
Correct! This will be done in https://github.com/zeit/micro/pull/219 once we have "micro" and "micro-dev" up and running. But for now, it's a good solution. Thanks!

PR #228, opened 2017-04-11
Fix server.listen error handling

dimapaloskin:
In the current implementation error handler for server.listen can never be called.

Part of `http` documentation concerns `callback` attribute for `listen` method:
> This function is asynchronous. callback will be added as a listener for the **'listening'** event. 

So we need to catch errors through 'error' event listener.


leo:
Thanks! 🙏

PR #241, opened 2017-04-24
Typedefinitions for the use with typescript

kaoDev:
with this declaration file the micro lib can be used with typescript.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,21 @@
+// Native
+import { IncomingMessage, ServerResponse } from 'http';
+
+type RequestHandler = (req: IncomingMessage, res: ServerResponse) => any
+
+export const run: (req: IncomingMessage, res: ServerResponse, fn: RequestHandler) => Promise<void>
+
+declare const serve: (fn: RequestHandler) => Promise<void>
+export default serve;
+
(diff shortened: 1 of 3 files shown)


leo:
Thank you for your effort on this! But please ask the team regarding such additions before you start working on them. In this case, for example, @zeit is not using TypeScript at all, which means that we won't be able to maintain this code. In addition, we don't have any plans on using it anywhere.

I hope you understand... 🙈

floatdrop:
@kaoDev you can add these to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) instead, which is more preferred way to add typing to node modules.

kaoDev:
@floatdrop the next stop will be the DefinitelyTyped repo, although I think the better way and also preferred way is to have the typings directly in the npm lib instead of having to install another dependency from the @types namespace.

@leo I wrote this definitions at first for my own TypeScript project, so the work was done anyway.

ythecombinator:
Hey @kaoDev! Thanks for your effort; I'm very interested in using your definitions but couldn't find them on DefinitelyTyped. Any plans?

kaoDev:
@ythecombinator sorry for the wait, I had just too much to do the last days. I will push an hour of opensource in today :)

PR #253, opened 2017-05-06
🐛 Export `micro` default

albinekb:
The readme says it should already be exported as default, however I found out that this is not true:

```
const server = micro_1.default((req, res) => __awaiter(this, void 0, void 0, function* () {
                              ^
TypeError: micro_1.default is not a function
```

This PR fixes that 👌


floatdrop:
@albinekb isn't your transpiler should do this for modules, that use CommonJS exports?

timneutkens:
When using babel `import micro from 'micro'` should work without this change 🤔

albinekb:
Hmm. This was with typescript, the `tsc` command with these settings:

```
{
  "compilerOptions": {
    "target": "ES6",
    "module": "commonjs",
    "lib": ["es2015", "es2017", "dom"],
    "jsx": "react",          
    "moduleResolution": "node",
    "typeRoots": ["./typings", "./node_modules/@types"],
    "allowSyntheticDefaultImports": true,
    "rootDir": "./src",
    "outDir": "./build"
  }
}
```

Without this change you need to include it like so:
```
import micro = require('micro')
```

With this change it's more "intuitive":
```
import micro from 'micro'
```


What's the downside of adding this `default` alias?

Maybe I just have the wrong typescript settings 😄

TooTallNate:
👍

albinekb:
But is this a correct change? @TooTallNate 🤔  or is typescript ~shitty~ not working like I expect it to?/my config wrong? 😄

albinekb:
Either way I can't see any downsides to adding this line so 👍  Go ahead and merge/close as you wish

TooTallNate:
Well we've definitely [done it in other projects already](https://github.com/zeit/async-retry/commit/64731d675822ed8f25950eff672070222195651b).

I can't really comment on the TypeScript config part because I'm not familiar enough with those options.

TooTallNate:
Also the reason that it works in babel is because it generates an `interopRequire()` function (or something like that) that first checks `exports.default` and then falls back to `module.exports` for that module IIRC.

timneutkens:
👌

albinekb:
😎

Issue #234, opened 2017-04-15
Micro is not very "lightweight"


styfle:
In the readme it says "micro" and "lightweight", but doesn't say what the installed size is. So I went ahead and did my own tests.

```sh
mkdir micro-example
cd micro-example
npm init -y
npm install --save micro
du -sh node_modules
```

This shows **9.2M** for `micro@7.3.2` and all it's dependencies.

Now lets try with the popular express framework.

```sh
mkdir express-example
cd express-example
npm init -y
npm install --save express
du -sh node_modules
```

This shows **1.8M** for `express@4.15.2` and all it's dependencies.

I also repeated the same steps for `koa@2.2.0` (**2M**) and `hapi@16.1.1` (**1.3M**) which are also popular web frameworks.

That means micro is the largest of the frameworks one might use to make microservices.

Is this expected?

leo:
`micro` is not just a framework, but also comes with a development CLI (express and koa don't, I don't know hapi). But yes, we're already working on splitting it out!

styfle:
> we're already working on splitting it out!

Splitting what out?

leo:
The development CLI... 😄

albinekb:
The code running is a lot less than express / koa, who cares about installed space in 2017?

leo:
@albinekb Yea, that's my opinion. But sadly, a bigger package also means longer installation times in production (which makes deployments slow).

styfle:
@albinekb My assumption is that `micro` was designed for implementing microservices.

If you add `micro` as a dependency to a 1000MB mono-repo, then it's only a 1% increase.

If you add `micro` as a dependency to 500 individual microservices, that's a 5000MB (5GB) increase in size.

austinfrey:
I'm getting 17mb now. @leo what's the status of splitting out the cli?

timneutkens:
@aafrey https://github.com/zeit/micro/pull/260

austinfrey:
@timneutkens thanks! what are you using to measure package size? the printout looks really nice!

timneutkens:
@aafrey this one https://github.com/siddharthkp/cost-of-modules

austinfrey:
thanks!

leo:
Very close!

tim-phillips:
The split happened: 

https://zeit.co/blog/micro-8

styfle:
I just checked v9.0.0 and it reports **747K**.
[...]

PR #260, opened 2017-05-14
Smaller bundle size

timneutkens:
Redo of #219

old:

screen shot 2017-05-14 at 14 37 32

new:

screen shot 2017-05-14 at 14 38 15

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,41 +1,33 @@
-// Packages
-const getPort = require('get-port')
-
 // Ours
 const serve = require('./server')
 const listening = require('./listening')
 const getModule = require('./module')
 
-module.exports = async (file, flags, module = getModule(file)) => {
+module.exports = (file, flags, module = getModule(file)) => {
   const server = serve(module)
 
-  let port = flags.port
-  let host = flags.host
+  const { port } = flags
+  const host = flags.host === '0.0.0.0' ? null : flags.host
 
-  const open = await getPort(port)
-  let inUse = open !== port
+  server
+    .listen(port, host, err => {
+      if (err) {
+        console.error('micro:', err.stack)
 
-  if (inUse) {
-    port = open
+        // eslint-disable-next-line unicorn/no-process-exit
+        process.exit(1)
+      }
 
-    inUse = {
-      old: flags.port,
-      open
-    }
-  }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -61,19 +61,13 @@
...
-    "chalk": "1.1.3",
-    "clipboardy": "1.1.2",
-    "get-port": "3.1.0",
-    "ip": "1.1.5",
     "is-async-supported": "1.2.0",
     "isstream": "0.1.2",
     "media-typer": "0.3.0",
     "node-version": "1.0.0",
     "raw-body": "2.2.0",
-    "update-notifier": "2.1.0"
```
albinekb, line 77: no more updates 🤔

Kikobeats: probably better create `micro-cli` and move inside

albinekb: very true 👌

Kikobeats:
looks better in terms of size, but you are removing the UX using micro from terminal, right?

lib/index.js
```diff
@@ -1,41 +1,33 @@
...
+      if (err) {
+        console.error('micro:', err.stack)
+
+        // eslint-disable-next-line unicorn/no-process-exit
+        process.exit(1)
+      }
+
+      return listening(server, host, flags.silent)
+    })
+    .on('error', err => {
```
albinekb, line 23: Should this be before `.listen`? 🤔

package.json
timneutkens: `micro-dev` is going to include it for both.

iamstarkov:
@timneutkens what is the reasoning behind this PR?

leo:
I'm closing this PR now, since we've began working on a huge rewrite of `micro` internally. However, this does not mean that this was done for nothing. Like I discussed it with @timneutkens, we'll be taking advantage of the pre-thinking he did and also credit him for that.

Please keep an eye on the releases page! 😊 🚀

styfle:
@iamstarkov This might be related to #234

iamstarkov:
@styfle thanks

iamstarkov:
@leo can you create an issue for rewrite, so we can track the progress and dont bother you in the meantime

leo:
This won't take long - the new `micro` will have the same API.

PR #263, opened 2017-05-19
Add Related section

Kikobeats:
what do you think about add a related section that connects with tiny modules that extend micro functionality?


leo:
Check out this list please 😋 https://github.com/amio/awesome-micro/blob/master/README.md

Kikobeats:
Maybe link awesome in the repo? 😂

Thanks!

Issue #224, opened 2017-04-07
HTTPS support?


AutoSponge:
Currently, [`http` is hard-coded](https://github.com/zeit/micro/blob/master/lib/server.js#L2). If you were to add `https` support, how would you prefer it done (cli flag, separate server.js, config file, package.json prop, etc.)?

albinekb:
Is it performant enough to run SSL with node? I've always been told to set up SSL via nginx proxy.

jamo:
My first intuition was that maybe this would be out of the scope from the most microscopic server. but then it could be really easy to support it 🤔 

But indeed for production it's in general recommended to do ssl with nginx and use it as a reverse proxy.

AutoSponge:
@albinekb I agree on performance, but in some hosting situations it may not be possible or cost-effective. It may also be helpful when using micro as a mock. [Fastify](https://github.com/fastify/fastify/blob/master/fastify.js#L52) provides an options object to solve this so I wasn't sure *IF* it were to add this feature, how micro would approach it.

dawsbot:
@AutoSponge http is not hardcoded, I deployed an https service with ease using micro and now [here](https://github.com/dawsbot/lat-lng). Am I misunderstanding the situation?

AutoSponge:
@dawsbot I'm pretty sure the now container translates https for you. For instance, you can't spin that up locally and hit it with an https:// curl, correct?

floatdrop:
It would be nice to support not only `http`/`https` modules, but accept any implementation of HTTP-server (like [uWebSockets](https://github.com/uWebSockets/uWebSockets)).

albinekb:
Indeed @AutoSponge,  @dawsbot: Now will set up an (nginx) reverse proxy and set up the SSL (https) for you. It's not done in micro

albinekb:
Actually it's really easy to do https with micro 🎉 #264

We could refactor a bit and add `micro-https` or similar to `bin` and it wouldn't impact the size/code being run at all. Thoughts? @jamo @AutoSponge  @floatdrop

leo:
Please check out the example that @albinekb wrote. We'll most likely not consider adding a special binary for HTTPs support, since it can be done like @albinekb suggested it. 😊

PR #264, opened 2017-05-20
Added HTTPS example

albinekb:
Just had a look at the source, and it's actually really easy to enable https! 🎉 

Since we want to keep micro **micro** I think the cli version of this should be published as a plugin to micro or as a separate cli that uses micro under the hood `micro-https`, thoughts?

That could use `ssl-self-signed-certificate` or similar to provision easy ssl for localhost/hosts-file domains when developing.

https://github.com/zeit/micro/issues/224

I'll publish this as a module so it can be used in zeit/serve if you want 🙌

### DIFF
--- a/examples/with-https/index.js
+++ b/examples/with-https/index.js
@@ -1,0 +1,21 @@
+const https = require('https')
+const { run, send } = require('micro')
+
+const cert = require('openssl-self-signed-certificate')
+
+const PORT = process.env.PORT || 3443
+
+const options = {
+  key: cert.key,
+  cert: cert.cert,
+  passphrase: cert.passphrase
+}
+
+const microHttps = fn => https.createServer(options, (req, res) => run(req, res, fn))
+
+const server = microHttps(async (req, res) => {
+  send(res, 200, { encrypted: req.client.encrypted })
+})
+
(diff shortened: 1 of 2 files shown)


examples/with-https/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+const PORT = process.env.PORT || 3443
+
+const options = {
+  key: cert.key,
+  cert: cert.cert,
+  passphrase: cert.passphrase
+}
+
+const microHttps = fn => https.createServer(options, (req, res) => run(req, res, fn))
```
albinekb, line 14: https://github.com/zeit/micro/blob/master/lib/server.js#L13

rauchg:
This could be cool for `micro-dev` perhaps? cc @timneutkens 

`micro-dev --ssl` ;)

Elyx0:
This is really important as otherwise, we get the `Mixed Content: This request has been blocked; the content must be served over HTTPS.` error if the frontend is *https* and the micro server *http*.

albinekb:
@Elyx0 this shouldn't (you can, but shouldn't) really be used in production though.

leo:
Thank you @albinekb! 😊

PR #284, opened 2017-07-20
Drain current requests before exit

randallsquared:
The close call never gets a chance to finish up awaiting requests before node exits, so those are dropped on the floor.


timneutkens:
That's a really good catch!

timneutkens (approved):

leo:
Thank you! 😊 https://github.com/zeit/micro/commit/facbef3032de4cfc499454c4c2e2c5cdd3d28f1d

PR #286, opened 2017-07-24
Minimize the core

leo:
This PR makes micro much tinier and even a little faster while the API stays the same.

## Installation Size

This is the dependency footprint that it leaves on the disk after this pull request:

screen shot 2017-07-24 at 11 30 02

How it was before:

screen shot 2017-07-24 at 11 43 46

## Performance

On my quad-core i5 3.2Ghz iMac, the benchmarks for requests per second for 10k requests ranged around **3,580 before and 4,400 after this PR** using this command:

```bash
ab -n 10000 http://localhost:4000/
```

## The Idea

As @rauchg and I discussed it internally, it will now do all of the basic thinking for serving a microservice in production and allowing people to test the production environment locally by running the command. However, all of the heavy development tools have been removed completely and will be added to micro-dev. In turn, the `micro` core will be lighter and faster!

After spending some time with minimist, nanomist and args (and others), I came to the conclusion that [mri](https://github.com/lukeed/mri) is the best and fastest one of all, so that's what I'm implementing everywhere now (it has basically the same API as minimist, but is much lighter). This makes `micro` even faster!


lib/index.js
```diff
@@ -1,41 +1,156 @@
...
+}
+
+exports.send = send
+exports.sendError = sendError
+exports.createError = createError
+
+exports.run = (req, res, fn) =>
+  new Promise(resolve => resolve(fn(req, res)))
+    .then(val => {
+      if (val === null) {
```
amio, line 98: Shall we check if response was sent (`res.headersSent`) already? To gracefully warning for use cases like this:
```javascript
const { send } = require('micro')
module.exports = async (req, res) => {
    send(res, 200, 'OK')  // in some where
    return 'Another OK'  // somehow
}
```

leo: Can you open an issue for this? To me, it sounds like a good idea - but I'd like to hear the opinion of the community. Not sure if there are special use cases. 😊

matheuss: Sounds like a nice idea for me too, but it sounds like a `micro-dev` feature – not sure how to _externalize_ such feature tho 🤔

timneutkens: @matheuss depending on how much of my ideas Leo has used it should be super easy to only support it in micro-dev 🙌

timneutkens: I'm +1 on this btw 👍

bin/micro.js
```diff
@@ -2,48 +2,50 @@
...
-      p: 'port',
-      H: 'host',
-      s: 'silent'
-    },
-    boolean: ['silent'],
-    string: ['host']
+// Check if the user defined any options
+const flags = parseArgs(process.argv.slice(2), {
+  string: ['host', 'port'],
+  boolean: ['help', 'version'],
```
lukeed, line 20: These will be `boolean` by default.

bin/micro.js
```diff
@@ -2,48 +2,50 @@
...
-    alias: {
-      p: 'port',
-      H: 'host',
-      s: 'silent'
-    },
-    boolean: ['silent'],
-    string: ['host']
+// Check if the user defined any options
+const flags = parseArgs(process.argv.slice(2), {
+  string: ['host', 'port'],
```
lukeed, line 19: You can set `default`s for these two & the type will be inferred; aka:

```js
default: {
  host: 'localhost',
  port: 3000
}
```

Will also apply the defaults & type-casting to their aliases.

lib/index.js
amio: Talks goes on here https://github.com/zeit/micro/issues/292 😃

j-f1:
What did you use to generate the tables in the first comment @leo?

leo:
`npm install cost-of-modules -g`

PR #298, opened 2017-08-17
Show hint, if error is about async on older Node.JS

floatdrop:
Not sure about `/\s+async\s+/.test(err.stack)` part, but it should be sufficient.

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -13,16 +13,18 @@ module.exports = file => {
     }
   } catch (err) {
     logError(`Error when importing ${file}: ${err.stack}`, 'invalid-entry')
+
     if (
       err instanceof SyntaxError &&
       /\s+async\s+/.test(err.stack) &&
       Number(process.versions.node.split('.')[0]) < 8
     ) {
       logError(
-        'Hint: In order for `async` & `await` to work, you need to use at least Node.js 8!',
+        'In order for `async` & `await` to work, you need to use at least Node.js 8!',
         'old-node-version'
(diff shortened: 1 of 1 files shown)


lib/handler.js
```diff
@@ -25,15 +12,17 @@ module.exports = file => {
       mod = mod.default
     }
   } catch (err) {
-    if (!checkAsyncAwait()) {
+    logError(`Error when importing ${file}: ${err.stack}`, 'invalid-entry')
```
leo, line 15: Newline after this please

lib/handler.js
```diff
@@ -25,15 +12,17 @@ module.exports = file => {
...
+    if (
+      err instanceof SyntaxError &&
+      /\s+async\s+/.test(err.stack) &&
+      Number(process.versions.node.split('.')[0]) < 8
+    ) {
       logError(
-        'In order for `async` & `await` to work, you need to use at least Node.js 8!',
+        'Hint: In order for `async` & `await` to work, you need to use at least Node.js 8!',
```
leo: Remove the "Hint:", please. The utility is printing its own prefix.

leo (changes requested):

lib/handler.js
floatdrop: Done.

lib/handler.js
floatdrop: Done.

leo:
Thank you!

PR #299, opened 2017-08-24
Added support for abstract streams

mikeal:
An instanceof check alone isn't quite enough for stream validation because it could also be an abstract stream conforming to the readable stream API.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,10 +1,11 @@
 // Native
 const server = require('http').Server
-const { readable } = require('is-stream')
+const { Stream } = require('stream')
 
 // Packages
 const getRawBody = require('raw-body')
 const typer = require('media-typer')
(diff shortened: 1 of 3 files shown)


timneutkens (approved):
Interesting! Looks 👌

leo:
@mikeal Looks wonderful! 😊 

Would you mind adding a test for this so that our coverage stays at 100%?

floatdrop:
@leo @timneutkens maybe it is better to pick up [is-stream](https://github.com/sindresorhus/is-stream) module for this?

leo:
Now taking advantage of `is-stream`. 🚀 

Released in [8.0.4](https://github.com/zeit/micro/releases/tag/8.0.4). Thank you, @mikeal! ❤️

PR #302, opened 2017-08-26
Set encoding to utf-8 for JSON response

### LINKED ISSUE #301: Shouldn't micro automatically set the charset of the response to UTF-8?

floatdrop:
Part of #301

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -69,7 +69,7 @@ const send = (res, code, obj = null) => {
     }
 
     if (!res.getHeader('Content-Type')) {
-      res.setHeader('Content-Type', 'application/json; encoding=utf-8')
+      res.setHeader('Content-Type', 'application/json; charset=utf-8')
     }
   }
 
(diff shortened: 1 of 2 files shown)


floatdrop:
@leo bump.

lib/index.js
```diff
@@ -69,7 +69,7 @@ const send = (res, code, obj = null) => {
     }
 
     if (!res.getHeader('Content-Type')) {
-      res.setHeader('Content-Type', 'application/json')
+      res.setHeader('Content-Type', 'application/json; encoding=utf-8')
```
ssut: I think this should be `charset=utf-8` instead of `encoding=utf-8` by [RFC2068](https://www.ietf.org/rfc/rfc2068.txt) and [the official w3 article](https://www.w3.org/International/articles/http-charset/index).

floatdrop: Oh, right! Thanks!

leo:
@floatdrop Thank you!

PR #313, opened 2017-09-24
ci: `.yml` file should be written in proper `.yml` syntax

luftywiranda13:
Make `.yml` really `.yml` 😬
It was json, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
-{
-  "language": "node_js",
-  "node_js": "node",
-  "after_success": [
-    "npx nyc report --reporter=text-lcov | npx coveralls"
-  ]
-}
+language: node_js
+node_js:
+  - 'node'
+after_success:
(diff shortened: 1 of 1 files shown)


timneutkens:
I believe this was intentional by @leo 🤔

leo:
YAML is a superset of JSON 😊 This is a JS project, so we need to use the JavaScriptObjectNationSyntax.

luftywiranda13:
what's wrong with using proper YAML, btw?
or is there any particular reason you stick using json in .yml? 🤔

leo:
@luftywiranda13 Travis CI doesn't support .json 😊

dimapaloskin:
@leo I guess @luftywiranda13 mean other stuff.
It's concerns file extension and their contents.
So when you make something like:
```
cat ./.travis.yml
``` 
you expect to see yaml syntax :)

PR #315, opened 2017-10-19
Remove console logging error stacks

### LINKED ISSUE #305: Remove console logging error stacks
After writing a couple of tests to confirm the right error handling paths are taken for my module my terminal started filling up with more and more logged errors by micro. Not sure what the intention is of logging these errors... Any way to disable them?

Found that commenting out https://github.com/zeit/micro/blob/master/lib/index.js#L84 is a workaround for now.

As a developer if I want to log errors I would handle that myself.

Siilwyn:
Logging errors here without the user having control over it doesn't seem like the best place.

Resolves #305


leo:
Thank you for the effort you've put into this, but I don't think we want that. By default, people should be able to see errors logged by micro in production... 😊

Siilwyn:
@leo After looking at this again I understand the logging, the linked issue is about logging during development though. Will put more details there.

PR #316, opened 2017-10-20
Replace media-typer module with content-type

dougwilson:
So this PR replaces the `media-typer` module with the `content-type` module. Both of these modules are part of the `jshttp` organization, and were doing similar things. This caused us to re-evaluate what exactly their roles were and so the `content-type` module is the one that parses the Content-Type HTTP header, which is what you are using it for. It works just like the version of `media-typer` you are currently using. The `media-typer` module will be releasing a new major that parses different strings, and so figured I'd help you move your framework now 👍

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,8 +3,8 @@ const server = require('http').Server
 const { Stream } = require('stream')
 
 // Packages
+const contentType = require('content-type')
 const getRawBody = require('raw-body')
-const typer = require('media-typer')
 const { readable } = require('is-stream')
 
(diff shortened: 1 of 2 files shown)


leo:
wow, thank's a lot! 😊 ❤️

dougwilson:
Forgot to mention that since I moved the code from `media-typer` to `content-type`, there are some performance improvements in there, so code that flows though that parse will likely be a bit quicker, especially for types without parameters on then (i.e. `application/json`), which are the most common to encounter.

leo:
Wonderful!

PR #320, opened 2017-11-02
Add port type and range validation to micro.

rapzo:
Assuming that unix sockets are not supported since the message identifies them as `undefined`.
I'll gladly patch [micro-dev](https://github.com/zeit/micro-dev) with the same, if the change makes sense.

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -82,6 +82,15 @@ if (!existsSync(file)) {
   process.exit(1)
 }
 
+const { port, host } = flags
+if (isNaN(port) || (!isNaN(port) && (port < 1 || port >= 2 ** 16))) {
+  logError(
+    `Port option must be a number. Supplied: ${port}`,
+    'invalid-server-port'
+  )
+  process.exit(1)
+}
+
 const loadedModule = handle(file)
 const server = serve(loadedModule)
 
@@ -90,7 +99,7 @@ server.on('error', err => {
   process.exit(1)
 })
 
-server.listen(flags.port, flags.host, () => {
+server.listen(port, host, () => {
   const details = server.address()
 
   process.on('SIGTERM', () => {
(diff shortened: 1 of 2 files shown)


leo:
Great idea, thanks!

bin/micro.js
```diff
@@ -82,6 +82,15 @@ if (!existsSync(file)) {
   process.exit(1)
 }
 
+const { port, host } = flags
+if (isNaN(port) || (!isNaN(port) && (port < 1 || port >= 2 ** 16))) {
```
sindresorhus, line 86: You should prefer `Number.isNaN` over `isNaN`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

leo:
@rapzo ^

rapzo:
On it. Thanks @sindresorhus!

bin/micro.js
frolic: The `2 ** 16` statement is causing trouble for me on Node ~~8.9.1~~ 6.5.0.
```
if (isNaN(port) || (!isNaN(port) && (port < 1 || port >= 2 ** 16))) {
                                                            ^
SyntaxError: Unexpected token *
```
Any reason to use this over `Math.pow` or just `port > 65535`?

**Edit:** `nvm` was misbehaving and running this under Node 6.5.0, but it works fine under 8.9.1. Feel free to ignore the above :)

revolunet: shouldn't we specify an `engines` field in the package.json so we at least get a npm warning when installing ?

rapzo: Nothing against `Math.pow`, just sugar syntax there. And since micro expects the usage of `async/await` i thought something that was available earlier wouldn't hurt.

https://github.com/zeit/micro/blob/2408dd57a2c2e4072563d7be0967652deeddb0bd/lib/handler.js#L23

revolunet: for some reason the error message was not displayed in my case with node v 6.2.3

upgraded to 8.7 and 👌

capture d ecran 2017-11-14 a 00 54 03

PR #322, opened 2017-11-03
Use `Number.isNaN` instead of `global.isNaN`

rapzo:
As pointed out in https://github.com/zeit/micro/pull/320#pullrequestreview-74197363


leo:
Wonderful. Thank you! 😊

PR #327, opened 2017-11-19
detect error parameter instance type and console accordingly

### LINKED ISSUE #314: Support non-Error instances error ?
`throw 'an error happened'`

In this case, micro doesn't print any logs except `undefined`. Should we handle this case, or we can make it clear you always have to throw `Error`.

arbazsiddiqui:
resolves #314


leo:
Thank you!

PR #335, opened 2018-01-07
Make it possible to export a promise that will eventually be a function

### LINKED ISSUE #317: Async actions before the server starts
There are a number of async actions you might want to do before the server starts. While this works fine if you create the listener itself, it precludes using things like `micro-dev` or the `micro` binary with this package. It'd be nice if we could include some async "setup" steps somehow. Maybe something like this?

```javascript
const db = new DB()
module.exports = async (req, res) => await db.getSomeAsyncData()
module.exports.setup = async () => { await db.connect() }
module.exports.teardown = async () => { await db.disconnect() }
```

EDIT: added teardown

timneutkens:
Fixes #317

This makes it possible to set up / initialize code before running the server. If we take this proposal we also have to update micro-dev.

Example:

```js


const initializeDatabase = require('some-database') // Purely illustrational

async function setup(fn) {
  const database = await initializeDatabase()

  return (req, res) => {
    return fn(req, res, database)
  }
}

async function handler(req, res, database) {
  return database.get('user')
}

module.exports = setup(handler)

```


sergiodxa:
This would make easier to integrate micro with Next.js as a custom server, 👍

leo:
Seems like a *very* good idea to me! 👏

sergiodxa:
@timneutkens I have a doubt, how this will work if I don't want to initialize anything? I need to export `async () => async () => { /* something here */ }`?

floatdrop:
@sergiodxa this should be backwards compatible – `await` on old handler function will return same handler function, so just export your handler function.

timneutkens:
Like @floatdrop said it's backwards compatible 👍
What this does is expect either a `function` or a `Promise` as the export.
So if you want to do async initialisation you can do the snippet I've included in the PR description. The `setup` function is executed when loading the file, meaning it is only execute once on startup. 👍

jamo (approved):

sijad:
how should it be used with es6 export default?

e.g. `export default setup(handler)`
I get `micro: https://err.sh/micro/no-export` error

timneutkens:
@sijad We've merged this PR https://github.com/zeit/micro/pull/336. It'll be supported in the next release.

aegatlin:
At the time of my writing this, the current version on npm is  `9.3.4`, which seems to include this commit. However, I can't get the following test file to work with zeit/nextjs. The file is at `pages/api/test.ts`:

```typescript
const test = async () => {
  return async (req, res) => {
    res.json({ hi: 'hey' })
  }
}

export default test()
```

If you remove the top level async, everything works, add it in, it breaks. Am I misunderstanding what this feature is supposed to be?

I'm running dev mode, so just `next` on the command line, so I would not be surprised if this has something to do with `micro-dev`, but I don't understand the relationships between them well enough to know for certain.

timneutkens:
Next.js does not use micro. They're separate projects.

aegatlin:
Thank you for your reply, but I'm not sure what to do now, and would appreciate guidance if you wouldn't mind.

The Nextjs documentation says it supports micro middleware, https://nextjs.org/docs/api-routes/api-middlewares, but I guess youre saying Nextjs won't support this type of behavior? Though, if it does support micro middleware, maybe it can somehow? Is there an alternative within nextjs?

(For context, my fundamental motivation is one-time data fetching before the server starts [to introspect a remote GraphQL schema]. I want to use top level await to do so, but nextjs doesn't seem to support it when compiling typescript due to syntax errors, thus I was hoping I could utilize this functionality within micro to wrap the await in an async function)

PR #336, opened 2018-01-12
[WIP] feat: support es6 module default export

tungv:
this allows projects that compiled through babel can conveniently `export default` a Promise

```js
// instead of
module.exports = app.prepare().then(() => service);

// babel users can do this
export default app.prepare().then(() => service);
```

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -9,7 +9,7 @@ module.exports = async file => {
     mod = await require(file) // Await to support exporting Promises
 
     if (mod && typeof mod === 'object') {
-      mod = mod.default
+      mod = await mod.default // Await to support es6 module's default export
     }
   } catch (err) {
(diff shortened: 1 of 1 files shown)


leo:
Wonderful addition! Can you add tests?

tungv:
@leo I'm not sure which test file is in charge of this kind of functionality currently. Probably I need to add tests for `handler.js`

tungv:
@leo I'm not sure how to test this, can you point me to the test for async setup function so I can have a clue?

timneutkens (approved):
This is perfect 👌

timneutkens:
@tungv I'll create an issue for the tests. I'll add them 👍

tungv:
thank you @timneutkens. As a heavy babel user, I'm longing for this feature

PR #345, opened 2018-03-07
Create snapshot test for CLI help

fmiras:
This PR adds snapshot testing to CLI help the same way as [serve tests it](https://github.com/zeit/serve/blob/master/test/options.js)


leo (approved):

leo:
Thank you so much! 😊

Issue #338, opened 2018-01-19
Create tests for Promise export


timneutkens:
Will add them whenever I have time to 👍 https://github.com/zeit/micro/pull/336

timneutkens:
This was fixed.

PR #346, opened 2018-03-08
Create handler test file and test promise cases

fmiras:
This PR solves #338. We should test also that handler exits process with code 1 when non-valid objects are received or if a non-function or object value is received at all.


timneutkens:
Please make sure the coverage stays the same 🙏

package-lock.json
```diff
@@ -80,6 +80,21 @@
         "arrify": "1.0.1"
```
timneutkens: lets remove the package-lock.json

timneutkens (approved):
Looks great, thank you @fmiras!

PR #347, opened 2018-03-10
Splat listen() args to only specify host when supplied

Qix-:
This more strictly abides by the contract that a non-specified host in the `listen()` call might behave differently if the system is running IPv4 vs IPv6 as per the documentation (since the documentation doesn't specify _exactly_ what happens on an IPv4 system with `'::'`).

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -17,7 +17,6 @@ const logError = require('../lib/error')
 // Check if the user defined any options
 const flags = parseArgs(process.argv.slice(2), {
   default: {
-    host: '::',
     port: 3000
   },
   alias: {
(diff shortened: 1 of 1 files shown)


bin/micro.js
```diff
@@ -101,7 +100,12 @@ async function start() {
     process.exit(1)
   })
 
-  server.listen(flags.port, flags.host, () => {
+  const listenArgs = [flags.port || 0]
```
TooTallNate, line 103: How would `port` be falsy if we default it to 3000?

Qix-: @TooTallNate empty string. I expected the option parser to accept an empty string as 'valid' and not use it as an excuse to fallback to default (where, instead, it should only populate the default if the `-p` option was never encountered at all).

Qix-: e.g. `micro -p ''`

TooTallNate (approved):
LGTM

PR #349, opened 2018-04-06
Avoid printing errors in test

### LINKED ISSUE #329: Suppress console errors on testing
Could it be possible to not dump errors to console while in test env?  It makes a total mess of tests. I have to compose all my handlers with a try/catch to swallow the dump. My error is tested through AVA throws but my screen gets filled with production error dumps.

stephenmathieson:
This patch prevents us from logging errors when testing (assumes NODE_ENV is set to "test" when testing).

Closes #329.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -83,6 +83,7 @@ const sendError = (req, res, errorObj) => {
   send(res, statusCode || 500, DEV ? errorObj.stack : message)
 
   // Avoid printing errors to the console when testing.
+  // istanbul ignore if
   if (process.env.NODE_ENV === 'test') {
     return
   }
--- a/lib/index.js
+++ b/lib/index.js
@@ -81,6 +81,12 @@ const sendError = (req, res, errorObj) => {
   const statusCode = errorObj.statusCode || errorObj.status
   const message = statusCode ? errorObj.message : 'Internal Server Error'
   send(res, statusCode || 500, DEV ? errorObj.stack : message)
+
+  // Avoid printing errors to the console when testing.
+  if (process.env.NODE_ENV === 'test') {
+    return
+  }
+
   if (errorObj instanceof Error) {
     console.error(errorObj.stack)
   } else {


Siilwyn:
@stephenmathieson I believe the kind of tests that trigger this path shouldn't exist, see my experience: https://github.com/zeit/micro/issues/305#issuecomment-345795261

stephenmathieson:
I disagree, and have a business requirement that all codepaths that are "exposed to the internet" are tested.

Siilwyn:
You're talking about 'category b' then right? This would be one log, which is manageable.

*Edit:* you could also handle all cases so that everything falls under 'category a' so nothing will be logged.

stephenmathieson:
Yeah, it's just one log per test/request, but it's annoying when you've got several tests hitting the unhappy path(s).

Siilwyn:
It should only be one test for the same path (the unhandled one where `micro` returns an error by default), right?

stephenmathieson:
Yes, but there are several paths in the function that will result in an error. Modified source follows:

```js
const assert = require('http-assert')

const app = micro(async (req, res) => {
  // We only care about POST requests.
  assert(req.method === 'POST', 405)

  // Ensure request is valid.
  const event = await json(req)
  assert(isValidWebhookRequest(event), 400)

  // Do other things.
  // ...

  // Finally, do the thing we care about.
  const meta = await do_business_logic({
    // ...
  })

  // Return the created metadata to the caller.
  // This may be useful later on. ¯\_(ツ)_/¯
  return meta
})
```

Siilwyn:
You can put a `try catch` around the functions that may throw [like this](https://github.com/Siilwyn/cheers-receiver/commit/a26a78ac2c03e43d19a5b1d4fbcb40cb6d4811a7#diff-e6a5b42b2f7a26c840607370aed5301aR15).

stephenmathieson:
Yeah, I could, but I'd rather not since Micro does it for me and just about every other framework handles these errors without printing to stdout.

Siilwyn:
Hmm I get that, I'm not fan of the logging either. On the other hand as I said before I think these paths shouldn't be tested since as you said the framework handles it.

Btw. check out my PR on this matter, one of the core contributors replied there: https://github.com/zeit/micro/pull/315

stephenmathieson:
I'm not testing that the framework handles the error, but instead, that my app raises the error.

_EDIT_: looking at your PR, I agree with Leo. Disabling logging entirely isn't what I want either; I just don't want noisy tests.

timneutkens:
This will be solved by #369, as there is no longer a "createError" handler in that scenario.

PR #351, opened 2018-04-15
Remove unnecessary async/await from readme

Zertz:


timneutkens:
This is actually there for demonstration purposes 👍

PR #353, opened 2018-04-23
Add support for `--unix-socket`

timneutkens:
As per @rauchg's request, added `--unix-socket` which will bind to a unix socket.

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -152,7 +152,7 @@ async function start() {
 
   const shutdown = () => {
     console.log('Gracefully shutting down')
-    server.close(() => process.exit())
+    server.close()
   }
   process.on('SIGINT', shutdown)
   process.on('SIGTERM', shutdown)
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -130,10 +130,13 @@ async function start() {
   server.listen(...listenArgs, () => {
     const details = server.address()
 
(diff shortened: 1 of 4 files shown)


bin/micro.js
```diff
@@ -16,12 +16,10 @@ const logError = require('../lib/error')
 
 // Check if the user defined any options
 const flags = parseArgs(process.argv.slice(2), {
-  default: {
-    port: 3000
-  },
   alias: {
     p: 'port',
     H: 'host',
+    s: 'socket',
```
nkzawa: We might call it `unix-socket` to make it clear? which is `curl`'s option name.

nkzawa (approved):

bin/micro.js
timneutkens: 👍 good call

bin/micro.js
```diff
@@ -115,8 +137,25 @@ async function start() {
...
+      console.log(`micro: Accepting connections on port ${details.port}`)
+      return
+    }
+
+    console.log('micro: Accepting connections')
   })
+
+  const shutdown = () => {
+    console.log('Gracefully shutting down')
+    server.close()
```
nkzawa: It seems there is another handler listening `SIGTERM` 🤔

nkzawa (approved):

leo (approved):

PR #356, opened 2018-04-24
Allow listening on multiple endpoints

Qix-:
This is technically a minor-compatible bump, but we might want to consider a major release.

This adds multi-listen support, where you can listen in multiple places at once on the same server handler.

This also changes the command line a bit (whilst deprecating old arguments - hence the "technically a minor bump" comment above). See the new section in `## Usage` in the readme for what that looks like.


timneutkens (approved):
💯

leo (approved):

styfle:
There was a small bump in install size after adding these dependencies.

v9.2.0 [install size](https://packagephobia.now.sh/result?p=micro@9.2.0)
v9.3.0 [install size](https://packagephobia.now.sh/result?p=micro@9.3.0)

But it's still looking way better than v7.x 👍 

v7.3.3 [install size](https://packagephobia.now.sh/result?p=micro@7.3.3)

Would you like me to add this badge to the README?

PR #358, opened 2018-04-26
Added hint about the port to example

dmitriz:
Just adding a helpful hint that everything works and where things are ;)

### DIFF
--- a/examples/socket.io-chat-app/index.js
+++ b/examples/socket.io-chat-app/index.js
@@ -15,4 +15,5 @@ const io = require('socket.io')(server);
 // socket-io handlers are in websocket-server.js
 require('./websocket-server.js')(io);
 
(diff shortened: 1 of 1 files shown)


examples/socket.io-chat-app/index.js
```diff
@@ -15,4 +15,5 @@ const io = require('socket.io')(server);
 // socket-io handlers are in websocket-server.js
 require('./websocket-server.js')(io);
 
+console.log('Listening on localhost:4000');
 server.listen(4000);
```
maximeshr: I think you should use the listen callback to do that.

```
server.listen(4000, () => console.log(':4000');
```

dmitriz: True, that way is better, thanks!

leo:
Thank you!

PR #367, opened 2018-06-30
Listen to process.env.PORT || 3000

brunolemos:
Fix heroku support.

### DIFF
--- a/bin/micro.js
+++ b/bin/micro.js
@@ -142,7 +142,7 @@ if (args['--port'] || args['--host'] || args['--unix-socket']) {
 
 if (args['--listen'].length === 0) {
 	// default endpoint
-	args['--listen'].push([3000]);
+	args['--listen'].push([process.env.PORT || 3000]);
 }
 
 let file = args._[0];


sergiodxa (approved):

styfle (approved):

timneutkens:
Can't you just do `micro -l $PORT` 🤔

timneutkens:
Essentially the same PR we closed because of the same reason: https://github.com/zeit/micro/pull/120

brunolemos:
> Can't you just do micro -l $PORT

No, got this error on heroku when using `micro -l $PORT`: 
`Invalid URL: 25103`

This works: `micro -l tcp://0.0.0.0:$PORT`
But it's not ideal IMO, it should just work. It's such a simple thing.

timneutkens:
It's non-standard behavior from Heroku's side, so it shouldn't "just work", instead we should document to use `micro -l tcp://0.0.0.0:$PORT` which we did before in the readme, but it was lost somewhere, I guess when `-p` was replaced with `-l`.

notrab (approved):

sunsean:
Does adding default detection of `$PORT` cause any problems?
Currently [cross-env-shell](https://www.npmjs.com/package/cross-env) or similar is required to write a run command that runs cross-os.

sunsean (approved):

timneutkens:
I'm going to close this PR as per my earlier comment. It's unexpected behavior + potentially breaking if `PORT` is defined in the current environment.

PR #375, opened 2018-10-24
convert getURL fn to more ergonomic one liner

coetry:
I'm a junior dev looking to start contributing to open source. I know this is a just a modest syntax adjustment and I'm not quite sure if this fits your style, but it doesn't break any tests and it saves a few bytes 😁


timneutkens (approved):
looks good to me

PR #381, opened 2018-12-11
remove unused argument for sendError

### LINKED ISSUE #343: Why req arg is required in sendError function
`type sendError = (req, res, errorObj) => void`

The function not read any values from `req`.

maxbeatty:
fixes #343 

feels too simple. what is being overlooked?

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ server.listen(3000)
 - Returns a [`http.Server`](https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_class_http_server) that uses the provided `function` as the request handler.
 - The supplied function is run with `await`. So it can be `async`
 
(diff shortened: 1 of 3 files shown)


timneutkens:
It's breaking / backwards incompatible right 🤔

maxbeatty:
absolutely. are you trying to keep everything backwards compatible? if not, is there a v10 milestone/timeline?

timneutkens:
As far as I'm aware this would indeed be a case where we wanted it to be backwards compatible as there are a massive amount of microservices out there that use micro. Even counting in ZEIT alone.

maxbeatty:
No worries. Completely understand. Should this and #343 be closed then?

timneutkens:
Thanks @maxbeatty 🙏 I've closed the issue.

Issue #369, opened 2018-08-01
RFC: Micro Response Type


rauchg:
The idea is to introduce an "envelope"-type construct into micro, that would allow us to express a broader range of HTTP responses as a simple `return` statement.

It restores a concept absent in the core Node.js HTTP API: a request comes in (function argument), a response goes out (`return`).

## Summary

- Introduces a `micro/res` return type
- Discourages imperatively responding, specially out-of-band (e.g.: callbacks)
- Deprecates `statusCode` from thrown errors. 
- Deprecates `micro/send` as an imperative wrapper ala `res.end`
- Adds official support of `TypeScript` / `Flow` types exports
- Makes micro even smaller!

### Respond with 404

```js
const res = require('micro/res')
module.exports = req => {
  return res('Not Found', 404)
}
```

### Respond with 204

```js
const res = require('micro/res')
module.exports = req => {
  return res(null)
}
```

### Respond with a JSON body

```js
const res = require('micro/res')
module.exports = req => {
  return res({ "hi": "there" })
}
```

### Respond with a stream

```js
const res = require('micro/res')
module.exports = req => {
  return res(fs.createReadStream('./photo.jpg'))
}
```

### Respond with JSON and headers

```js
const res = require('micro/res')
module.exports = req => {
  return res({ error: "not_found" }, 404, {
     'Access-Allow-Control-Origin': '*'
  })
}
```

### Fun Trivia: Symmetry

If we avoid `{` and `}`, it almost looks like we're using the imperative `res` (similar to our current `send`):

```js
const res = require('micro/res')
module.exports = req => (
  await sideEffect(),
  res('Your request has been logged', 202)
)
```

### Composition

The return value from `res(200)` would return a `Response` object that works like an `immutable-js` `Record`.

I've written before about why [*middleware as a framework construct is a bad idea*](https://github.com/zeit/micro/issues/8#issuecomment-178362486), because functional programming is the ultimate middleware framework. Our language already has all the tools we need.

So let's imagine we want to create _"middleware"_ that always sends CORS headers to all your responses. The end goal would be to use it as follows:

```js
module.exports = withCORS(req => 
  res(200, 'Hi there')
)
```

So, how do we write `withCORS`?

```js
async function withCORS(fn) {
  const res_ = await fn()
  return res_.setHeaders({
    'Access-Control-Allow-Origin': '*'
  })
}
```

Which can actually be simplified as 

```js
async function withCORS(fn) {
  return (await fn()).setHeaders({
[...]

PR #382, opened 2018-12-18
Typescript migration

alisabzevari:
Related to #369,

**This is a WIP PR. Do not merge but add as much comments as you can.**

**TODO:**
- [x] Fix code coverage
- [ ] Convert tests to Typescript
- [ ] Decide about the CLI. Should it be written in Typescript?
- [ ] Fix eslint and add tslint (if needed)
- [ ] Finalize project structure
- [ ] Add more strict Typescript compilation parameters
- [ ] Fix all TODOs in the code

**Note:**

The goal of this PR is to convert the project to Typescript without breaking backward compatibility. Ideally, we should be able to release a new minor version with proper typing information. This will help us to refactor the project to the new proposed change (Micro response type).


src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+export type RequestHandler = (
+	req: IncomingMessage,
+	res: ServerResponse
+) => ResponseObject | undefined;
+
+export const serve = (fn: RequestHandler) =>
+	new Server((req, res) => run(req, res, fn)); // TODO: We should create an issue in DefinitelyTypes: `Server` can be called without new
+
+export default serve;
```
timneutkens: This is a breaking change, as `const micro = require('micro')` will hold an object that has `.default`. I'd change it back to `module.exports = serve`

alisabzevari: I fixed it. Tests are now exactly the same as before.

alisabzevari:
I am closing this PR in favor of #383.

PR #383, opened 2018-12-27
Fix #369 - New Response Type

### LINKED ISSUE #369: RFC: Micro Response Type

alisabzevari:
To resolve #369,

I've mixed the implementation of the new response type with typescript migration. Also changed the testing framework from ava to jest because jest had better support for typescript.


.vscode/launch.json
```diff
@@ -0,0 +1,32 @@
+{
```
rauchg: I'd like to avoid editor config in source control

rapzo (commented):
Why the name change from handler to handle? Shouldn't it be as close as possible?

alisabzevari:
Before this PR, `handler` was a default export (sort of) which means the name doesn't matter. People have to find a name when they import that module and it was named as `handle` when imported in the cli script. But your question is important and is related to this question which is still unanswered in my mind:
* Should we export something as default? (IMO, default exports are evil. They make es6 and commonjs modules incompatible and make refactoring hard as well. So, I didn't export anything as default just because development was easier with them. I will change them to make them similar to the current version if you think we should.)

alisabzevari:
Summary of the last changes:
* Rexported all the public functionality of library through index.ts. This shows that the project doesn't have any default exports, but I couldn't convince myself to have a default export expect backward compatibility. See [here](https://github.com/zeit/micro/pull/383/commits/17f95385885c9b85775c5024e0b73e2e2c7b9f08).
* I have removed the concept of `statusCode` in errors in the whole project. I hope I understood @rauchg’s comments correctly. Please look at [this commit](https://github.com/zeit/micro/pull/383/commits/7bdf44df284ef4d1dd20872562730ad7f4798859) for this part of the changes.
* Removed unused packages and updated other packages. Some packages added their own type definitions which is safer to use them instead of DefinitelyTyped. Also, I have removed packages regarding commit hooks etc. I will add them later once I figured out a good set of linting roles and tools.
* Enabled compiler `strict` switch and I could defeat the compiler expect in `cli.ts`. It would be great if you could help me fix them. I have added some comments in this PR regarding them.
* Removed old js files from the project.

src/cli.ts
```diff
@@ -1,42 +1,45 @@
...
+	"--port": Number,
+	"-p": "--port",
+	"--host": String,
+	"-h": "--host",
+	"--unix-socket": String,
+	"-s": "--unix-socket"
+};
+
+// Check if the user defined any options
+const args = arg(argSpec as any);
```
alisabzevari: This shouldn't be `any`.

src/cli.ts
```diff
@@ -189,16 +198,16 @@ function registerShutdown(fn) {
...
-	process.on('SIGTERM', wrapper);
-	process.on('exit', wrapper);
+	process.on("SIGINT", wrapper);
+	process.on("SIGTERM", wrapper);
+	process.on("exit", wrapper);
 }
 
-function startEndpoint(module, endpoint) {
-	const server = serve(module);
+function startEndpoint(module: HttpHandler, endpoint: any) {
```
alisabzevari: This shouldn't be `any`.

.gitignore
```diff
@@ -8,3 +9,9 @@ yarn-error.log
 # coverage
 coverage
 .nyc_output
+
+# output
+lib
```
dav-is, line 14: How come we are ignoring the output now?

src/error.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function err(
+	message?: string,
+	originalError?: Error,
+	statusCode?: number
+): HttpError {
+	return new HttpError(message, originalError, statusCode);
+}
+
+// export interface Exception extends Error {
```
dav-is: Needs to be removed, would remove it myself unless it was reminding you to do something

dav-is:
I tested these changes on the examples and found two issues:
`json-body-parsing` example doesn't seem to work anymore. Says `invalid JSON`
`urlencoded-body-parsing` is returning an empty `{}`
`external-api-call` is fetching a URL that doesn't exist

We need an update to the README too, right?

I also feel as if we need some CLI testing and we could easily get to (or close to) 100% coverage

.gitignore
alisabzevari: The output (`lib` folder) should be used for only publishing. I would prefer to ignore it because then we have to make sure to not commit the stale output. It is basically the compiled result.

alisabzevari:
> json-body-parsing example doesn't seem to work anymore. Says invalid JSON
urlencoded-body-parsing is returning an empty {}
external-api-call is fetching a URL that doesn't exist

What are the expected results? I will then write tests for them to cover them.

dav-is:
@alisabzevari I'm not really sure what the expected results are... doesn't seem to be acting any differently than before this PR, so if you don't want to investigate it and make them easier to understand/function, then I won't make it a blocker to merging this PR.

Does this section of the README need edits? https://github.com/zeit/micro#sending-a-different-status-code and https://github.com/zeit/micro#error-handling

Maybe we should add the fact that `micro` is now using Typescript in the features?

alisabzevari:
Summary of last changes:
* Added a new function called `listener`. This function only abstracts the `requestListener` that should be passed to `Server`. This function will be useful in creating http servers. See `with-https` example.
* Updated all the examples to work with the new api. We can see how the breaking changes look like.
* I think it makes sense to add another example using typescript.

alisabzevari:
I've changed the code in a way to be exactly backward compatible + the new `res` function. So all the old tests are now working exactly like before. I just migrated them to jest and added some more to cover the new `res` function.

### Notes
* `res` is not correctly importable with `micro/res`. We have to fix this to comply with RFC.
* We have to add some examples with Typescript and the new `res` function.
* More tests needed to cover `cli.ts`.
* We have to add deprecation warnings to the functions that we want to depreciate. I need some opinions on this to know how to do it?

dav-is:
> We have to add deprecation warnings to the functions that we want to depreciate. I need some opinions on this to know how to do it?

You can use `./error`'s `logError()` and add an error page to `/errors`
https://github.com/zeit/micro/blob/007171a418483ea933edc0a2f3fdcb0788bdd8c5/src/error.ts#L1

Might want to implement a way to use `console.warn` instead of `console.error`.

alisabzevari:
Summary of last changes:

* Added deprecation warnings to `send` and `sendError`. They will show up only in dev mode.
* Added two error descriptions for deprecation errors and wrote some workarounds and samples on how to fix them.
* I figured out that [DefinitelyTyped contains some typing information for micro](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/micro). I have changed the names of the types and did some minor changes in types to match typing info in DT.
* Updated some packages.
* Added documentation to readme describing the new `res` function.
* Added type information for every function mentioned in the readme to better document the API.

alisabzevari:
Is there any plans regarding this PR?

PR #387, opened 2019-01-12
Automatically set Content-Length 0 if 204 status

mfix22:
Automatically sets the `Content-Length` header for when sending `204`'s, since Safari expects it (as seen here: https://github.com/expressjs/cors/blob/master/lib/index.js#L176-L177).

If you don't want to support this, no worries — feel free to close 🙂

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -29,6 +29,10 @@ const send = (res, code, obj = null) => {
 	res.statusCode = code;
 
 	if (obj === null) {
+		if (code === 204 && typeof res.setHeader === 'function') {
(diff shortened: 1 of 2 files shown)


iamstarkov:
looks good to me

iamstarkov:
wut

iamstarkov:
@mfix22 why?

mfix22:
@iamstarkov I'm actually not so sure this is actually necessary anymore. 
1. I think Safari no longer has this issue
2.Not _certain_ the maintainers  we need so pollute this package with edge cases
3. I figured I could always reopen if necessary 🤷‍♂️ 

(i'll reopen)

PR #389, opened 2019-02-26
Automatically set Content-Length 0 if 204 status

mfix22:


timneutkens:
@mfix22 can you add a test for this

mfix22:
@timneutkens after looking into this further, I don't think this should be included in this library. User's can easily perform this action themselves.

timneutkens:
Okay let's close this then.

PR #395, opened 2019-05-02
Decrease install size

timneutkens:
- Remove below Node 8 check because node 6 is end-of-life
- Remove chalk as it was only used in `--help`
- Upgrade `arg` as the newer version is about 80% smaller.


styfle (dismissed):
Amazing! 🎉

PR #397, opened 2019-05-02
Bring in is-stream

timneutkens:
This slightly reduces bundle / install size because only one function of is-stream was being used.


styfle (approved):
Nice! Shaved off 6KB

PR #399, opened 2019-05-11
Add new programmatic usage

timneutkens:
Related: https://github.com/zeit/next.js/issues/7297

For serverless functions, `@now/node`:

```js
const micro = require('micro')

module.exports = micro((req, res) ⇒ {
   return 'Hello World'
})
```

Current invoking micro() will return a Node.js http.Server which doesn't work well with just `req, res` as a handler. This also ties into using micro standalone with `@now/node`. Doing this will also allow us to deprecate `micro-dev` too in favor of using Next.js / `now dev` + `@now/node`

The new programmatic usage works everywhere and really makes Micro a layer on top of the default Node.js request handling.

This solves:

- Easier usage (no extra steps needed besides installing and using, like setting up the commands in package.json etc is no longer needed)
- Easier deployment
- Scalable by default
- Serverless by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,13 +244,14 @@ module.exports = async (req, res) => {
 You can use Micro programmatically by requiring Micro directly:
 
 ```js
+const http = require('http')
 const micro = require('micro')
 const sleep = require('then-sleep')
 
-const server = micro(async (req, res) => {
+const server = new http.Server(micro(async (req, res) => {
   await sleep(500)
   return 'Hello world'
-})
+}))
 
 server.listen(3000)
 ```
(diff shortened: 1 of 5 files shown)


javivelasco (approved):
I think it could be a good idea to add the `index.d.ts` file with proper types so we can stop using `@types/micro` and there will be no need to update there either.

timneutkens:
Sounds great to me, but out of scope for this PR 👍

rauchg (approved):

matheuss (approved):

zartdinov:
Looks like need to update README.md.

https://github.com/zeit/micro#programmatic-use
https://github.com/zeit/micro#testing

I had some problems with using `http.Server`

timneutkens:
Actually no, the readme is correct. It's just not released on stable yet. You can check https://npmjs.com/micro for the stable readme.

zartdinov:
I understood, thanks!

PR #411, opened 2019-07-22
Bump `raw-body` version

thasophearak:
This PR bump `raw-body` package which contain `iconv-lite` version that have a fix for

```
[DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

### DIFF
--- a/packages/micro/package.json
+++ b/packages/micro/package.json
@@ -27,6 +27,6 @@
   "dependencies": {
     "arg": "4.1.0",
     "content-type": "1.0.4",
-    "raw-body": "2.3.2"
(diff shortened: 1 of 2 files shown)


timneutkens (changes requested):
Seems like this is not needed if it's merged on canary already.

timneutkens (approved):
Actually as you pointed out in slack this has to be on `master` and not the `canary` branch, because that one has to be removed.

PR #421, opened 2019-11-29
Add ZEIT Now Disclaimer

mcsdevv:
This PR adds a disclaimer to the repository, advising users that Micro is not required when using ZEIT Now. The disclaimer replaces the redundant warning about the canary branch.


packages/micro/README.md
```diff
@@ -1,4 +1,4 @@
-**The below readme is the documentation for the `canary` (prerelease) branch. To view the documentation for the latest stable Micro version visit [npmjs.com/micro](https://npmjs.com/micro).**
```
timneutkens, line 1: This should be preserved.

timneutkens (approved):

PR #427, opened 2020-02-06
use master branch for is-stream license link

mariuz:


mariuz:
Maybe project url is better ? https://github.com/sindresorhus/is-stream/

timneutkens:
The current link is correct, it was based on that exact version of the code.

PR #428, opened 2020-02-11
Use vanilla JS for DOM manipulation instead of JQuery

kresnasatya:
Hi, I found the example of socket.io-chat-app using jQuery for DOM manipulation and I thought it would be better to use vanilla JS instead of JQuery for DOM. Thank you.


OlliV:
FYI: I picked your commit to here https://github.com/OlliV/micri/commit/9a3204b3a4eb9273e19bf9503528b727923ba355

leerob (approved):
Thank you! (Sorry for the delay)

PR #443, opened 2021-08-31
Update "iconv-lite" to v0.6.3

TooTallNate:
There is an issue with "iconv-lite" v0.4.x such that it invokes the deprecated `new Buffer()` function, which causes Node.js to print an ugly deprecation warning at runtime.

The `raw-body` dep is the actual consumer of the iconv module, however even its latest version still uses the problematic iconv version (there's [a stale PR](https://github.com/stream-utils/raw-body/pull/67) to update it). So switching to a fork that I published with only that version update change. Confirmed that this change makes the deprecation warning not be triggered.

### DIFF
--- a/packages/micro/lib/index.js
+++ b/packages/micro/lib/index.js
@@ -3,7 +3,7 @@ const {Stream} = require('stream');
 
 // Packages
 const contentType = require('content-type');
-const getRawBody = require('raw-body');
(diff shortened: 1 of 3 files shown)


TooTallNate:
I see now that the `canary` tag of micro has the latest `raw-body` dep already, which also has `iconv-lite` without the warning. So I'll close this one. We should publish a stable release though.

PR #448, opened 2021-11-04
Native ES modules import for .mjs

PonomareVlad:
Fixes this issue https://github.com/vercel/micro-dev/issues/87 by native-way

### DIFF
--- a/packages/micro/lib/handler.js
+++ b/packages/micro/lib/handler.js
@@ -5,7 +5,7 @@ module.exports = async file => {
 	let mod;
 
 	try {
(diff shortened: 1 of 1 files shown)


leerob (approved):
Thank you! Ideally we could have some tests for this 🙏

leerob:
Would you be interested in adding some here? https://github.com/vercel/micro/tree/master/test

leerob:
Turns out this is a bit harder to write a unit test for than I expected. If anyone has any good ideas, please let me know!

PR #458, opened 2022-07-17
Add ESLint and TypeScript

### LINKED ISSUE #457: Add ESLint and move to TypeScript
Copy setup from https://github.com/vercel/ms

pmbanugo:
Closes https://github.com/vercel/micro/issues/457

This PR adds TypeScript and rewrites the library code using TypeScript. 

**Highlight:**

- Rewrote the library using TS
- Rewrote the test using TS and node-tap
- Updated the GitHub Actions
- Generate types for the library
- Removes deprecated endpoints
- Drops Node.js 12 support


leerob:
Amazing! Will dive into this and take a look.

packages/micro/src/bin/micro.ts
```diff
@@ -0,0 +1,270 @@
...
+
+if ((args['--port'] || args['--host']) && args['--unix-socket']) {
+  logError(
+    `Both host/port and socket provided. You can only use one.`,
+    'invalid-port-socket',
+  );
+  process.exit(1);
+}
+
+// TODO: consider removing the deprecated endpoints support with a major release?
```
leerob: Yeah, this would be a good time to do it - this will be published as a major version.

leerob: We can also drop Node.js 12 support.

packages/micro/src/bin/micro.ts
```diff
@@ -0,0 +1,270 @@
...
+  server.listen(endpoint, () => {
+    const details = server.address();
+    registerShutdown(() => {
+      console.log('micro: Gracefully shutting down. Please wait...');
+      server.close();
+      process.exit();
+    });
+
+    // `micro` is designed to run only in production, so
+    // this message is perfectly for prod
```
leerob: "perfectly for prod" seems like a typo.

packages/micro/src/lib/handler.ts
```diff
@@ -0,0 +1,30 @@
+// Utilities
+import { logError } from './error';
+
+export const handle = async (file: string) => {
+  let mod: unknown;
+
+  try {
+    mod = file.endsWith('.mjs') ? await import(file) : await require(file); // Await to support exporting Promises
```
leerob: Heads up: I ended up reverting this from the other PR just to make sure we had time to test it out here before the major release.

pmbanugo: I've updated to match the latest main commit. I think it can be done in a different PR.

packages/micro/src/bin/micro.ts
pmbanugo: I've removed those and set 14.5 as the minimum Node version

test/development.js
```diff
@@ -1,44 +0,0 @@
-// Packages
```
pmbanugo, line 1: I didn't add a test for this because the library is only to be used in production. Therefore, I don't find any benefit in checking for `NODE_ENV === 'development'`

The same applies for **production.js**.

the function `sendError` is updated accordingly

packages/micro/src/lib/index.ts
```diff
@@ -0,0 +1,215 @@
...
+      res.setHeader('Content-Type', 'application/octet-stream');
+    }
+
+    res.setHeader('Content-Length', obj.length);
+    res.end(obj);
+    return;
+  }
+
+  if (obj instanceof Stream || readable(obj)) {
+    //TODO: Wouldn't (obj instanceof Stream) be the only check here? Do we specifically need a Readable stream or a Stream object that's not of NodeJS Stream?
```
pmbanugo, line 83: I'm not sure about this. The comment is for someone here to let me know why we do this. Or discuss later in the future if this is necessary.

pmbanugo:
Hi @leerob It's ready for review.

leerob:
I think we'll need to add a `test` script for CI.

CleanShot 2022-07-24 at 13 51 35@2x

pmbanugo:
> I think we'll need to add a `test` script for CI.
> 
> CleanShot 2022-07-24 at 13 51 35@2x

@leerob  I've updated the script, it should work now.

pmbanugo:
That should work but it's failing 😓 I'm not sure why, but I'll check again.

pmbanugo:
@leerob Updated... If it doesn't work this time, perhaps you can try clearing the cache and rerunning the build.

leerob:
It works! 🎉

leerob (approved):
Will try to get one more review on here tomorrow, but LGTM.

pmbanugo:
> Will try to get one more review on here tomorrow, but LGTM.

Hi, @leerob just following up on this 😄

leerob:
Planning to go ahead and merge/release https://github.com/vercel/micro/pull/460, and then let's release this as a major version 👍

pmbanugo:
@leerob Do you want me to update my branch and use the reverted change, or go ahead with what I currently have with `serve` and `http.Server`? I guess you would want the second option as part of the major version release.

leerob:
What do you think? I think we probably keep the breaking changes.

pmbanugo:
I think we keep the breaking change because that way we can use it programmatically with fastify or something similar. I'll make the necessary changes and update the PR soon.

pmbanugo:
I left it with the breaking change and updated the documentation

leerob (approved):

PR #460, opened 2022-07-29
Bring back default http server wrapper

walaszczykm:
The micro has a breaking change in its minor version upgrade from 9.3.4 to 9.4.0 related not to default wrapping the "serve" function result into `http.Server` instance, which leads to "server.listen is not a function" error.

The difference can be seen by comparing the source and README:
- https://www.npmjs.com/package/micro/v/9.3.4#programmatic-use
- https://www.npmjs.com/package/micro/v/9.4.0#programmatic-use

This PR proposes reverting this part of the code and readme to the state from version 9.3.4

### DIFF
--- a/packages/micro/README.md
+++ b/packages/micro/README.md
@@ -214,18 +214,15 @@ module.exports = async (req, res) => {
 You can use Micro programmatically by requiring Micro directly:
 
 ```js
-const http = require('http');
-const micro = require('micro');
-const sleep = require('then-sleep');
+const micro = require('micro')
+const sleep = require('then-sleep')
 
-const server = new http.Server(
-  micro(async (req, res) => {
-    await sleep(500);
-    return 'Hello world';
-  })
-);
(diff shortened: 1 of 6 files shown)


walaszczykm:
Hey @leerob, sorry for the confusion, I saw that tests are hanging ad dug a little bit into why. Turns out the `_test-utils.js` was wrapping micro into `new http.Server` instance. I have adjusted it accordingly.

I feel quite hesitantly as those are changes from 3 years ago, and now I am reverting them, but I see they all were all rolled out as a canary release until now, the new 9.4.0 version. If you feel that this new way of consuming micro programmatically is better suited, then maybe going with it in a dedicated major release would also be an option.

leerob (approved):

Bessonov:
From my point of view, it would be nice to get rid of `http.Server`. It makes micro usable with another implementations like `https.createServer`.

PR #468, opened 2022-09-16
fix typo in the body parsing section

IAmSSH:


leerob:
I think "an" is correct here!

IAmSSH:
I think "an" is used to refer to a single entity. Here since we are talking about exporting multiple functions (3, to be precise):

>... included **an** async **functions** `buffer`, `text` and `json`... 

So I thought maybe it is not required.

But anyways, thanks for reviewing 😄

PR #470, opened 2022-10-01
docs: update README imports

RyKilleen:
Update examples to reflect `serve` as the default export of `micro`


leerob (approved):

pmbanugo:
@leerob  I just want to mention that the current TS codebase doesn't have a default export which why the doc is written that way. It should be considered in case you decide to release a new version with the current code.

Issue #474, opened 2023-01-14
v10 broken on windows


t8g:
Running micro on windows fail with 

![image](https://user-images.githubusercontent.com/639759/212468495-baad8b22-6c72-4da5-a33b-7d0e5e6c3fba.png)

I think it's due to the usage of `import(file)` in `handler.ts`.
It should be fixed with the usage of `pathToFileURL` https://nodejs.org/api/url.html#urlpathtofileurlpath

```
import(pathToFileURL(file).href)
```

Thanks for your work on this package !

PR #475, opened 2023-01-14
Use pathToFileURL to handle windows file path

t8g:
Related to #474 

Fix `On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'`


leerob:
Could you add a test please?

t8g:
The current tests are already failing on windows.
And the PR make them pass.
But I don't know how to emulate the error on non windows plateform.

